PR #2, opened 2021-01-08
setup auto build for read the doc integration

eddyxu:


changhiskhan (approved):
lgtm. Only small nit is to promote the python code path to a module level variable.

PR #7, opened 2021-01-10
Improve numpy doc

eddyxu:
Add examples for `rikai.numpy` utilities and the purpose of them


changhiskhan (approved):
Maybe add a blurb about why being able to do this is useful?

PR #8, opened 2021-01-10
Rename Bbox to Box2d for consistency with 3d bounding box object

eddyxu:


eddyxu:
This should be reviewed after #4 .

PR #9, opened 2021-01-11
Rename BBox to Box2d

eddyxu:


changhiskhan (approved):
just confirming: this is just moving things around right? no logic changes really?

eddyxu:
> just confirming: this is just moving things around right? no logic changes really?

Mostly about renaming. But it also changed the bonding box def from `xmin, xmax, ymin, ymax` to `x, y, height, width` to make it consistent with the way `Box3d` is defined.

PR #13, opened 2021-01-13
[video] Add sampling capabilities

changhiskhan:
Create utilities for sampling videos using opencv then
integrate that into DataLoader to reduce boilerplate.
We use prefetch to sample videos in parallel but each video
is sampled by the same thread to keep the order of the frames.


python/rikai/torch/data.py
```diff
@@ -131,3 +135,29 @@ def __iter__(self) -> Generator[Dict, None, None]:
             yield self.collate_fn(batch)
         prefetch_thd.join()
         q.join()
+
+
+def make_video_sampler(sample_gen, video_column_name=None):
```
eddyxu, line 140: Add doc here? how should user expect to use it

python/rikai/video.py
```diff
@@ -0,0 +1,137 @@
...
+        :param max_samples: yield at most this many frames
+        """
+        super().__init__(stream)
+        self.sample_rate = sample_rate
+        self.start_frame = start_frame
+        self.max_samples = max_samples
+
+    def __iter__(self):
+        # TODO use seek for sparse sampling and maybe multithreaded
+        cap = cv2.VideoCapture(self.stream.uri)
```
eddyxu, line 99: do we need to put `cv2` into deps?

python/rikai/video.py
```diff
@@ -0,0 +1,137 @@
...
+class SingleFrameSampler(VideoSampler):
+    """
+    A simple sampler that just returns one out of every `sample_rate` frames
+    """
+
+    def __init__(
+        self, stream: VideoStream, sample_rate: int, start_frame: int, max_samples: int
+    ):
+        """
+        :param stream: source video to sample from
```
eddyxu, line 87: We should have a style guide :) 

i used [numpy docstring](https://numpydoc.readthedocs.io/en/latest/format.html) for the rest of the python codebase.  Lets get some consent on docstring format ?

python/rikai/video.py
```diff
@@ -0,0 +1,137 @@
+from abc import ABC, abstractmethod
```
eddyxu, line 1: Apache license. 

And rename this to `python/rikai/types/video.py`?

python/rikai/video.py
```diff
@@ -0,0 +1,137 @@
...
+import cv2
+import youtube_dl
+
+from rikai.spark.types import (
+    YouTubeVideoType,
+    VideoStreamType,
+)
+from rikai.vision import Image
```
eddyxu, line 9: Is this import needed?

python/rikai/video.py
```diff
@@ -0,0 +1,137 @@
...
+        from IPython.lib.display import YouTubeVideo
+
+        return YouTubeVideo(self.vid)._repr_html_()
+
+    def __eq__(self, other) -> bool:
+        return isinstance(other, YouTubeVideo) and self.vid == other.vid
+
+    def get_stream(self, **kwargs) -> "VideoStream":
+        # TODO allow configurable cache dir
+        dl_path = "/tmp/{}.mp4".format(self.vid)
```
eddyxu, line 34: use `TMPDIR`?

python/rikai/video.py
changhiskhan: These were auto-generated by IntelliJ. I prefer numpy docstring actually, I'll do a docs/comments cleanup pass

python/rikai/video.py
changhiskhan: Do you think the sampling classes should go in types or in like `python/rikai/utils.py` ?

python/rikai/video.py
changhiskhan: copy-pasta. i'll remove.

python/rikai/torch/data.py
changhiskhan: I will add it there. Current usage is in the notebook i added. I can move some code snippets over.

python/rikai/video.py
eddyxu: You can change intellij to set default docstring to numpy style.

python/rikai/video.py
eddyxu: should we have a `python/rikai/video.py` for video related functionalities?

i'd imagine that we will have

```
rikai.types -> for UDT
rikai.torch -> for torch access
rikai.featurestore (or .dataset) -> for dataset management
rikai.video(.sample/.other function) -> for video specific operations?

```

you know what, what about `rikai.transforms.video`?

python/rikai/video.py
changhiskhan: good idea. I will use TMPDIR for this PR. I need to make this more configurable and also automatically add an option to cp to s3 as well.

python/rikai/video.py
changhiskhan: added

python/rikai/video.py
changhiskhan: yeah, i had project settings set before but I created a new one for the new repo and lost it

python/rikai/video.py
changhiskhan: i put everything in rikai.types.video for now. We can decide where to put these utility functions.

python/__init__.py
```diff
@@ -0,0 +1,13 @@
+#  Copyright 2021 Rikai Authors
```
eddyxu, line 1: Do we need this level ? 🤣

eddyxu (approved):
lgtm

python/__init__.py
changhiskhan: I just have intellij automatically add it to all .py files lol

PR #18, opened 2021-01-15
Add Segment and some unit tests for video-related classes

changhiskhan:


python/rikai/spark/types/video.py
```diff
@@ -93,3 +94,41 @@ def deserialize(self, datum) -> "YouTubeVideo":
...
+        super().__init__()
+
+    def __repr__(self) -> str:
+        return "SegmentType"
+
+    @classmethod
+    def sqlType(cls) -> StructType:
+        return StructType(
+            fields=[
+                StructField("start_fno", IntegerType(), nullable=False),
```
eddyxu, line 112: It might be just my ignorance. what is `fno`? mind to add some doc?

python/rikai/spark/types/video.py
```diff
@@ -93,3 +94,41 @@ def deserialize(self, datum) -> "YouTubeVideo":
...
+                StructField("end_fno", IntegerType(), nullable=False),
+            ]
+        )
+
+    @classmethod
+    def module(cls) -> str:
+        return "rikai.spark.types"
```
eddyxu, line 119: It is not a big deal, but I'd lean to use the strictest module as possible (i.e. `rikai.spark.types.video`), i am worried that we will soon get into cycle-import hell.

python/rikai/types/video.py
```diff
@@ -150,6 +148,31 @@ def __eq__(self, other) -> bool:
         return isinstance(other, VideoStream) and self.uri == other.uri
 
 
+class Segment:
+    """A video segment bounded by frame numbers"""
+
+    __UDT__ = SegmentType()
+
+    def __init__(self, start_fno: int, end_fno: int):
+        if start_fno < 0:
```
eddyxu, line 157: This number is 0-indexed?
Mind to add some doc regarding to `fno`?

src/main/scala/org/apache/spark/sql/rikai/Video.scala
```diff
@@ -20,6 +20,49 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.UTF8String
 
+/**
+  * A reference to a particular Youtube vid. A given Youtube vid is associated with many
+  * content streams. These can be video streams of different formats at different bitrates,
+  * audio-only streams, and video-only streams.
+  *
+  * @param uri
```
eddyxu, line 28: The constructor uses `vid`, should we update this `@param uri`

python/rikai/spark/types/video.py
changhiskhan: that's fair. I'll change it.

python/rikai/spark/types/video.py
changhiskhan: frame number. I'm using fno to keep consistent with opencv terminology (the assumption is that, whether right or wrong, most ML people working with videos in python will use opencv)

eddyxu (approved):
LGTM overall.

The remained nits in the comments already.

changhiskhan:
> LGTM overall.
> 
> The remained nits in the comments already.

thanks. latest should have addressed your comments. will merge

PR #21, opened 2021-01-18
use uniformed VERSION file for both scala and python

eddyxu:
Use the same file to specify version for both scala (`sbt`) and python (`setup.py`).
It also fixes the `python setup.py build` on Mac using `sbt in homebrew`


python/setup.py
```diff
@@ -6,16 +6,18 @@
...
 class BuildCommand(distutils.command.build.build):
     def run(self):
         jars_dir = os.path.join("rikai", "jars")
         if os.path.exists(jars_dir):
             shutil.rmtree(jars_dir)
         os.makedirs(jars_dir)
-        sbt = "/usr/bin/sbt"
-        check_call([sbt, "clean"], cwd=os.pardir)
-        check_call([sbt, "package"], cwd=os.pardir)
+        check_call(["sbt clean"], cwd=os.pardir, shell=True)
```
eddyxu, line 19: `homebrew` installed `sbt` at `/usr/local/bin/sbt`.

changhiskhan (commented):
what do you think about introducing something like https://github.com/c4urself/bump2version/ to help manage semantic versioning?

eddyxu:
Fancy ;) sure, lets do it

PR #24, opened 2021-01-19
add quickstart draft and update index page

eddyxu:
Emphasis the main feature of rikai format.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -19,6 +19,7 @@ Lets get started from the feature engineering in `Spark`_.
 
 .. code-block:: python
 
+    from pyspark.sql import SparkSession
     from pyspark.ml.linalg import DenseMetrix
     from rikai.types import Image, Box2d
     from rikai import numpy as np
@@ -39,17 +40,17 @@ Lets get started from the feature engineering in `Spark`_.
                 "mat": DenseMatrix(2, 2, range(4)),
                 "image": Image("s3://foo/bar/1.png"),
                 "annotations": [
-                {
-                    "label": Label("cat"),
-                    "mask": np.random(size=(256,256)),
-                    "bbox": Box2d(x=1.0, y=2.0, width=3.0, height=4.0)
-                }
+                    Row(
+                        label=Label("cat"),
+                        mask=np.random(size=(256, 256)),
+                        bbox=Box2d(x=1.0, y=2.0, width=3.0, height=4.0)
+                    )
                 ]
             },
         ]
     )
 
-   df.write.format("rikai").save("s3://path/to/features")
+    df.write.format("rikai").save("dataset/out")
 
 We can then inspect the dataset in a `Jupyter Notebook`_.
 
@@ -57,7 +58,7 @@ We can then inspect the dataset in a `Jupyter Notebook`_.
 
 .. code-block:: python
 
-    df = spark.read.format("rikai").load("out")
+    df = spark.read.format("rikai").load("dataset/out")
     df.printSchema()
     df.show(5)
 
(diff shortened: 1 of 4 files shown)


docs/quickstart.rst
```diff
@@ -0,0 +1,83 @@
...
+        .master("local[*]")
+        .getOrCreate()
+    )
+
+    df = spark.createDataFrame(
+        [
+            {
+                "id": 1,
+                "mat": DenseMatrix(2, 2, range(4)),
+                "image": Image("s3://foo/bar/1.png"),
```
changhiskhan, line 40: can Image handle http reference? Maybe we can make this work with an actual image?

docs/quickstart.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Use `Rikai` in the machine learning life cycle
+----------------------------------------------
+
+Lets get started from the feature engineering in `Spark`_.
+
+
+.. code-block:: python
+
+    from pyspark.ml.linalg import DenseMetrix
```
changhiskhan, line 22: clarify that you need pyspark setup (and point to pyspark setup instructions otherwise) ?

docs/quickstart.rst
```diff
@@ -0,0 +1,83 @@
+.. toctree::
+   :maxdepth: 1
+
+Quickstart
```
changhiskhan, line 4: reference the docker image if people want to start with a bare environment and not install into their local system?

docs/quickstart.rst
```diff
@@ -0,0 +1,83 @@
...
+                    "label": Label("cat"),
+                    "mask": np.random(size=(256,256)),
+                    "bbox": Box2d(x=1.0, y=2.0, width=3.0, height=4.0)
+                }
+                ]
+            },
+        ]
+    )
+
+   df.write.format("rikai").save("s3://path/to/features")
```
changhiskhan, line 52: write to a local path instead?

docs/quickstart.rst
```diff
@@ -0,0 +1,83 @@
...
+
+We can then inspect the dataset in a `Jupyter Notebook`_.
+
+
+
+.. code-block:: python
+
+    df = spark.read.format("rikai").load("out")
```
changhiskhan, line 60: use the same local path as the previous example?

docs/quickstart.rst
```diff
@@ -0,0 +1,83 @@
...
+
+
+Use the dataset in `pytorch`
+
+.. code-block:: python
+
+    from rikai.torch import DataLoader
+
+    data_loader = DataLoader(
+        "s3://path/to/features",
```
changhiskhan, line 73: same here

changhiskhan (approved):
I'll add the video related things separately

docs/quickstart.rst
eddyxu: Do you have an actual image URL in mind?

docs/quickstart.rst
eddyxu: How does databricks deploy these jars? should we use `spark.jar.packages` or do we expect a deployment method that install rikai by default on cluster @changhiskhan

docs/quickstart.rst
eddyxu: done

docs/quickstart.rst
eddyxu: done

docs/quickstart.rst
eddyxu: done

docs/quickstart.rst
eddyxu: i will add this after we have a docker release procedure.

PR #30, opened 2021-01-22
Enable mac build

eddyxu:
Enable mac build, and makes the `setup.py` works with `sbt` from `homebrew`.


changhiskhan (approved):
you don't need to explicitly install numpy anymore?

eddyxu:
No. It was there for `fastparquet` setup's . We migrated to pyarrow 2.0 , it does not need numpy to run setup.py anymore.

PR #32, opened 2021-01-22
Improve docstrings

eddyxu:


python/rikai/types/vision.py
```diff
@@ -97,23 +103,3 @@ def __eq__(self, other) -> bool:
 
     def to_numpy(self) -> np.ndarray:
         return np.array([self.label])
-
-
-class Annotation(ToNumpy):
```
eddyxu, line 102: This class was ill-designed / half baked, and not used by anything yet.

Lets delete this for now, and find a better solution.

changhiskhan (approved):

PR #33, opened 2021-01-22
Cleanup/refactor for image/video types and test code

changhiskhan:
1. Add `.show(**kargs)` method to Image/YouTubeVideo/VideoStream.
This allows the user to pass in custom display options. Each of these
classes have default _repr_*_ display methods but `.show` allows for
custom options (like size)

2. Implement _repr_*_ methods for Image.
For local/embedded resources, _repr_html_ doesn't work.

3. Remove youtube download option for now.
We can just use pafy to get a single stream and then sample from that.
This removes a very heavy weight dependency with uncertain status

4. Remove __init__ files from test directories (py.test requirement)
Suggestion for testing convention: use py.test convention rather than
unittest/nose convention.


python/rikai/types/vision.py
```diff
@@ -40,12 +40,20 @@ def __init__(self, uri: str):
...
 
-        TODO: find more appropriate way to return image in jupyter?
-        """
-        return f"<img src={self.uri} />"
+        return Image(self.uri)._repr_mimebundle_(include=include, exclude=exclude)
+
+    def _repr_jpeg_(self):
+        from IPython.display import Image
+
+        return Image(self.uri)._repr_jpeg_()
```
eddyxu, line 51: Does this work with `s3://` `gs://` URIs?

python/rikai/types/vision.py
```diff
@@ -40,12 +40,20 @@ def __init__(self, uri: str):
     def __repr__(self) -> str:
         return f"Image(uri={self.uri})"
 
-    def _repr_html_(self):
-        """Jupyter integration
+    def _repr_mimebundle_(self, include=None, exclude=None):
```
eddyxu, line 43: mind to add doc here?

python/rikai/types/vision.py
```diff
@@ -40,12 +40,20 @@ def __init__(self, uri: str):
...
-        """
-        return f"<img src={self.uri} />"
+        return Image(self.uri)._repr_mimebundle_(include=include, exclude=exclude)
+
+    def _repr_jpeg_(self):
+        from IPython.display import Image
+
+        return Image(self.uri)._repr_jpeg_()
+
+    def _repr_png_(self):
```
eddyxu, line 53: How are we going to test this?

Not urgent, but what do you think that we extend these core POD class with juptyer in a separate module?

A little bit concern that we will accidentally drag heavy deps like Jupyter into production code.

python/rikai/types/vision.py
changhiskhan: let me test that (at least with s3)

python/rikai/types/vision.py
changhiskhan: ok will do

python/rikai/types/vision.py
changhiskhan: should be easy enough just to put a few small images in the test directory. and we can also add some webtests to refer to web image references (and s3/gs ones)

eddyxu: https://ipython.readthedocs.io/en/stable/config/integrating.html#formatters-for-third-party-types  

how about this.  This is basically MVC while these types are `M`, and the extensions are `V`

python/rikai/types/video.py
```diff
@@ -59,42 +58,53 @@ def __init__(self, vid: str):
     def __repr__(self) -> str:
         return "YouTubeVideo({0})".format(self.vid)
 
-    def _repr_html_(self):
-        from IPython.lib.display import YouTubeVideo
+    def show(self, **kwargs):
+        """
+        Visualization in jupyter notebook with custom options
+
+        Parameters
```
eddyxu, line 65: do you have an example about how to use it?  

Are these per-object level options? Can/should we config them globally?

python/rikai/types/video.py
```diff
@@ -59,42 +58,53 @@ def __init__(self, vid: str):
...
         Parameters
         ----------
-        download: bool, default False
-            By default just return a remote https reference. If True then
-            actually download the data
-        kwargs
+        ext: str, default 'mp4'
+            The preferred extension type to get
+        quality: str, default 'worst'
+            Either 'worst' or 'best'
```
eddyxu, line 95: worth to use an `Enum`?

python/rikai/types/video.py
```diff
@@ -59,42 +58,53 @@ def __init__(self, vid: str):
...
+
+    def _repr_html_(self):
+        """default visualization in jupyter notebook cell"""
+        return self.show()._repr_html_()
 
     def __eq__(self, other) -> bool:
         return isinstance(other, YouTubeVideo) and self.vid == other.vid
 
-    def get_stream(self, download: bool = False, **kwargs) -> "VideoStream":
+    def get_stream(self, ext="mp4", quality="worst") -> "VideoStream":
```
eddyxu, line 86: adding type hint to parameters?

python/rikai/types/vision.py
```diff
@@ -37,15 +37,41 @@ def __init__(self, uri: str):
         super().__init__(uri)
         self._cached_data = None
 
+    def show(self, **kwargs):
```
eddyxu, line 40: Hmm, do we expect `show()` to be implemented by all the object?  a little bit worried that this makes the jupyter integration not transparent.

python/tests/types/test_vision.py
```diff
@@ -0,0 +1,51 @@
...
+import pytest
+
+from binascii import b2a_base64
+from PIL import Image as PILImage
+import numpy as np
+
+from rikai.types.vision import Image
+
+def test_show_embedded_png(tmp_path):
+    data = np.random.random((100,100))
```
eddyxu, line 23: lol not like to use `unittest.TestCase`?

python/rikai/types/vision.py
changhiskhan:
```
The notebook and the Qt console can display richer representations of objects. To use this, you can define any of a number of _repr_*_() methods. Note that these are surrounded by single, not double underscores.
```

isn't that what we're doing?

I added some more tests for these methods just now.

Jupyter: is it that heavy compared to all of the other dependencies in the production system (pytorch/pandas/spark/numpy/etc)

python/rikai/types/video.py
changhiskhan: these are the kwargs for the IPython.display.YouTubeVideo/Video/Image classes. I can add links to their docs.

python/rikai/types/vision.py
eddyxu: my take is that the jupyter integration code is on a separate module to the actual data objects.
Which makes jupyter / display module a `trait` to extend existing codes.

python/rikai/types/vision.py
changhiskhan: the default _repr_*_ makes it transparent if you don't need additional display options. The `show` just allows for customization

python/tests/types/test_vision.py
changhiskhan: yeah, makes the code simpler most of the time, but I'm open to discussion on testing convention.

eddyxu: i am fine either way.  lets get a consistent style ?

python/rikai/types/video.py
changhiskhan: probably not rn. Let's let the API stablize first

python/rikai/types/video.py
changhiskhan: done

python/rikai/types/vision.py
changhiskhan: we just need the `ipython` module, not all of jupyter. It looks like the package is 1mb according to conda forge. At this stage in the project life-cycle, I would vote that we keep it together. wdyt?

.github/workflows/python.yml
```diff
@@ -34,6 +34,7 @@ jobs:
         python -m pip install -q numpy wheel pylint black
         python setup.py bdist_wheel
         python -m pip install $(ls dist/rikai*.whl)[test]
```
eddyxu, line 36: can we do `install ... [test,youtube]`

python/rikai/types/vision.py
eddyxu: should we rename `show()` to `display` ? it is consistent with jupyter's naming conventions?

It is like
```
#(inexplicably) 
display(video) 
==>
video.display(foo, bar)
```

python/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    webtest: marks tests as making a network call (deselect with '-m "not webtest"')
```
eddyxu, line 3: shall we use the standard test terms , like unit / integration / system / smoke?

eddyxu (approved):
LGTM overall. Small nits

.github/workflows/python.yml
changhiskhan: ok let me change before merge

PR #35, opened 2021-01-22
Remove analyze module (for now)

changhiskhan:
For the first release let's keep the functionality more limited
so that it's cleaner and easier for early users to understand and
use without errors.


eddyxu (approved):
SGTM

PR #44, opened 2021-01-23
Add pycodestyle check in github action and Makefile

eddyxu:
`make lint` now support both pycodestyle and black

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,8 +5,5 @@ exclude =
     venv,
     dist,
     build
-max-line-length = 99
 
 [pycodestyle]
-max-line-length = 99
-max-doc-length= 200
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,6 @@
+[flake8]
+exclude =
+    .git,
+    __pycache__,
+    venv
+max-line-length = 99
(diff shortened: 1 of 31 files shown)


eddyxu:
`flake8` seems to have much more warnings, we can do it later.

setup.cfg
```diff
@@ -0,0 +1,12 @@
...
+exclude =
+    .git,
+    __pycache__,
+    venv,
+    dist,
+    build
+max-line-length = 99
+
+[pycodestyle]
+max-line-length = 99
```
changhiskhan, line 11: 79 or 99?

setup.cfg
```diff
@@ -0,0 +1,12 @@
...
+max-doc-length= 200
```
changhiskhan, line 12: is this PEP8 compliant?

eddyxu: How to handle long URLs in doc? 

We can strictly follow with PEP8 and disable lint on such lines manually. 
What do you think?

setup.cfg
eddyxu: Deliberately change this to `99` from the default value (79).

I am fine to stick on `79`

setup.cfg
changhiskhan: using `noqa`? I think that's fine

changhiskhan (approved):
bombs away

Issue #45, opened 2021-01-24
Make a jupyter notebook working with coco dataset.


eddyxu:

eddyxu:
Closed by #47

PR #47, opened 2021-01-24
Coco notebook

eddyxu:
For #45


changhiskhan (approved):
1. Maybe add a note saying you need to publishLocal the jar first in cell 1
2. Add a comment saying the coco dataset download might take a while in cell 2

PR #50, opened 2021-01-24
Vision related UDFs

eddyxu:
Provide UDFs for easy operations on image and `box2d`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,6 +8,7 @@ scalaVersion := "2.12.11"
 
 libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-sql" % "3.0.1",
+  "io.netty" % "netty-transport-native-epoll" % "4.1.46.Final",
   "software.amazon.awssdk" % "s3" % "2.14.12",
   "com.google.cloud" % "google-cloud-storage" % "1.113.0",
   "org.scalatest" %% "scalatest-funsuite" % "3.2.0" % "test"
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,6 @@ scalaVersion := "2.12.11"
 
 libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-sql" % "3.0.1",
-  "io.netty" % "netty-transport-native-epoll" % "4.1.46.Final",
   "software.amazon.awssdk" % "s3" % "2.14.12",
   "com.google.cloud" % "google-cloud-storage" % "1.113.0",
   "org.scalatest" %% "scalatest-funsuite" % "3.2.0" % "test"
--- a/python/rikai/spark/functions/geometry.py
+++ b/python/rikai/spark/functions/geometry.py
@@ -22,13 +22,25 @@ from rikai.spark.types.geometry import Box2dType
 # Rikai
 from rikai.types.geometry import Box2d
 
-__all__ = ["area", "box2d"]
+__all__ = ["area", "box2d", "box2d_from_center"]
 
 
 @udf(returnType=Box2dType())
 def box2d(coords) -> Box2d:
     """Build a Box2d from ``[xmin,ymin,xmax,ymax]`` array."""
-    return Box2d(coords[0], coords[1], coords[2], coords[3])
+    return Box2d(*coords)
+
+
+@udf(returnType=Box2dType())
+def box2d_from_center(coords) -> Box2d:
+    """Build a Box2d from center-point based coordinates,
+    ``[center_x, center_y, width, height]`` array.
+
+    See Also
+    --------
+    :py:meth:`rikai.types.geometry.Box2d.from_center`
+    """
+    return Box2d.from_center(*coords)
 
 
 @udf(returnType=FloatType())
--- a/python/rikai/spark/functions/geometry.py
+++ b/python/rikai/spark/functions/geometry.py
@@ -52,7 +52,8 @@ def box2d_from_top_left(coords) -> Box2d:
     Example
     -------
 
-    `Coco dataset <https://cocodataset.org/>`_ is one public dataset that use top-left coordinates
+    `Coco dataset <https://cocodataset.org/>`_ is one public dataset that use
+    top-left coordinates
 
 
     >>> #! pyspark --packages ai.eto.rikai:rikai_2.12:0.0.1
--- a/python/rikai/spark/functions/geometry.py
+++ b/python/rikai/spark/functions/geometry.py
@@ -1,0 +1,30 @@
+#  Copyright 2021 Rikai Authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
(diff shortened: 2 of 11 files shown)


eddyxu:
This is supporting the image notebook demo

build.sbt
```diff
@@ -8,6 +8,7 @@ scalaVersion := "2.12.11"
 
 libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-sql" % "3.0.1",
+  "io.netty" % "netty-transport-native-epoll" % "4.1.46.Final",
```
changhiskhan, line 11: where is this required?

python/rikai/spark/functions/geometry.py
```diff
@@ -0,0 +1,98 @@
...
+from rikai.logging import logger
+from rikai.types.geometry import Box2d
+
+__all__ = ["area", "box2d", "box2d_from_center", "box2d_from_top_left"]
+
+
+@udf(returnType=Box2dType())
+def box2d(coords) -> Box2d:
+    """Build a Box2d from ``[xmin,ymin,xmax,ymax]`` array."""
+    return Box2d(coords[0], coords[1], coords[2], coords[3])
```
changhiskhan, line 32: nit: check xmax > xmin and ymax > ymin?

python/rikai/spark/functions/geometry.py
```diff
@@ -0,0 +1,98 @@
...
+
+# Rikai
+from rikai.logging import logger
+from rikai.types.geometry import Box2d
+
+__all__ = ["area", "box2d", "box2d_from_center", "box2d_from_top_left"]
+
+
+@udf(returnType=Box2dType())
+def box2d(coords) -> Box2d:
```
changhiskhan, line 30: So the user workflow you're imagining is like: 
1. the bounding boxes will come out of the training model as an numpy array
2. they store that in a column or a struct field
3. then they call this box2d udf on it to convert to Box2d?

python/tests/spark/test_functions.py
```diff
@@ -30,8 +30,8 @@ def test_areas(self):
         """Test calculating bounding box's area."""
         df = self.spark.createDataFrame(
             [
-                (Box2d(1, 2, 1.0, 1.0),),
-                (Box2d(10, 12, 1.0, 5.0),),
+                (Box2d(1, 2, 2.0, 3.0),),
```
changhiskhan, line 33: maybe add a couple of basic cases to test the udf?

python/rikai/types/geometry.py
```diff
@@ -60,67 +63,114 @@ def to_numpy(self) -> np.ndarray:
...
-        X-coordinate of the center point of the box
-    y : float
-        Y-coordinate of the center point of the box
-    width : float
-        The width of the box
-    height : float
-        The height of the box
+    xmin : float
+        X-coordinate of the top-left point of the box.
+    ymin : float
```
changhiskhan, line 72: the terminology is a bit confusing because top-left should be `ymax` no?

python/rikai/types/geometry.py
```diff
@@ -60,67 +63,114 @@ def to_numpy(self) -> np.ndarray:
...
-    def __init__(self, x: float, y: float, width: float, height: float):
+    def __init__(self, xmin: float, ymin: float, xmax: float, ymax: float):
+        self.xmin = float(xmin)
+        self.ymin = float(ymin)
+        self.xmax = float(xmax)
+        self.ymax = float(ymax)
+
+    @classmethod
+    def from_center(
+        cls, center_x: float, center_y: float, width: float, height: float
```
changhiskhan, line 90: nit: consider adding some basic validation. Since these constructors all have the same input types, sometimes users might use the wrong one and it'll succeed silently but give the wrong result. Adding some basic validation (e.g., width and height > 0) isn't foolproof but if the user uses the wrong one on a whole column these will catch the vast majority of human errors.

src/main/scala/org/apache/spark/sql/rikai/Box2d.scala
```diff
@@ -62,32 +64,32 @@ class Box2dType extends UserDefinedType[Box2d] {
   override def sqlType: DataType =
     StructType(
       Seq(
-        StructField("x", DoubleType, nullable = false),
-        StructField("y", DoubleType, nullable = false),
-        StructField("width", DoubleType, nullable = false),
-        StructField("height", DoubleType, nullable = false)
+        StructField("xmin", DoubleType, nullable = true),
```
changhiskhan, line 67: why nullable = true?

build.sbt
eddyxu: I will remove this. 

For some reason, `pyspark` does not work with the `sbt publishLocal` package.  I will fix them later

python/rikai/spark/functions/geometry.py
eddyxu: will do

python/rikai/spark/functions/geometry.py
eddyxu: Right, and for many open dataset (i.e., coco), bounding box are stored as list of numbers in json format.
Here is offering a UDF for conversion.

python/rikai/types/geometry.py
eddyxu: the top-left point is `(0,0)`

python/rikai/spark/functions/geometry.py
changhiskhan: sg that makes sense

python/rikai/types/geometry.py
changhiskhan: ah ok. you think most users will already know the coordinate system or would they need a docstring reminder?

eddyxu: while i know these are 3 most common coordinate systems, i am not sure the function is good tho.   open for suggestion.

python/rikai/types/geometry.py
eddyxu: Sure will do.

Will using `assert` break prod system tho?  Are there sth like `guava.precondition` that can be turn off in prod in python?

src/main/scala/org/apache/spark/sql/rikai/Box2d.scala
eddyxu: will revert.

python/tests/spark/test_functions.py
eddyxu: done

python/rikai/types/geometry.py
changhiskhan: no i meant just add a note in the docstring above saying something like 'note: this uses graphics coordinate system where the top-left is (0,0)'

python/rikai/types/geometry.py
changhiskhan: python has a -o flag

changhiskhan (approved):
lgtm. if you have time just add the documentation nit to reminder users about the coordinate system.

eddyxu:
Thanks @changhiskhan . added docstring for `(0, 0)`

Issue #27, opened 2021-01-20
Implement serialize and deserialize on NDArray in scala


eddyxu:
Without them, we can not run `df.show()` with `numpy.ndarray`

eddyxu:
Is [mxnet.NDArray](https://mxnet.apache.org/versions/1.7.0/api/scala/docs/tutorials/ndarray) too excessive to bring in `mxnet`?

what do you think, @changhiskhan

eddyxu:
Another alternative could be serialize `numpy` into a 1-d pyarrow array, and on the java / scala side, create a dimensional view on this array, this could be more performant, but i dont have good exposure to `scala` to see what API looks like

eddyxu:
I will take this one

eddyxu:
Closed by #51

PR #51, opened 2021-01-25
Numpy array display-able in Spark df.show()

eddyxu:
Print out a summary of `numpy.array` in a Spark DataFrame (#27)


python/tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@pytest.fixture(scope="session")
+def spark() -> SparkSession:
+    # TODO: load rikai jar from "sbt publishLocal" installed version?
+    return (
+        SparkSession.builder.appName("spark-test")
+        .config(
+            "spark.jars",
+            os.path.join(
```
changhiskhan, line 29: nit: consider using pathlib instead: 
```
from pathlib import Path

Path(rikai.__file__).parent() / 'jars' / 'rikai_2.12-0.0.1.jar'
```

python/tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+                os.path.dirname(rikai.__file__), "jars", "rikai_2.12-0.0.1.jar"
```
changhiskhan, line 30: is there an easy way to pull the current version number? If it's a hassle, just create a github issue and we can wait til after 0.0.1

src/main/scala/org/apache/spark/sql/rikai/NDArray.scala
```diff
@@ -17,10 +17,46 @@
...
+import org.apache.spark.sql.catalyst.InternalRow
 
+/**
+  * NDArray type. Storing the persisted data from numpy ndarray.
+  *
+  * Note that this class does not support interoperation in scala yet.
+  * It will only show a array summary in "df.show()".
+  * But the data is accessible from Pyspark and Pytorch / Tensorflow readers.
+  *
+  * @todo consider to use arrow Array for interchangable data transfer.
```
changhiskhan, line 30: I don't thnk we should invest more in scala interop for numpy ndarray. Let's wait til we're able to take advantage of more Arrow integrations.

src/main/scala/org/apache/spark/sql/rikai/NDArray.scala
```diff
@@ -17,10 +17,46 @@
...
+      case 2  => "uint16"
+      case 3  => "uint32"
+      case 4  => "uint64"
+      case 5  => "int8"
+      case 6  => "int16"
+      case 7  => "int32"
+      case 8  => "int64"
+      case 9  => "float"
+      case 10 => "double"
+      case 11 => "boolean"
```
changhiskhan, line 57: what if the ndarray isn't any of these types (https://numpy.org/devdocs/user/basics.types.html) or if it's a subtype (e.g., datetime64, timedelta64, etc)?

python/tests/conftest.py
eddyxu: we can use a `glob` actually.  Do you think we should wait for `sbt publishLocal` to install a `-SNAPSHOT` version?

src/main/scala/org/apache/spark/sql/rikai/NDArray.scala
eddyxu: agreed. so put this into `todo` :)

python/tests/conftest.py
eddyxu: I kind lean to not have this `jars` directory in the future.  Even on a Databricks cluster, the jars will be deployed separately from this `pip install rikai`.

src/main/scala/org/apache/spark/sql/rikai/NDArray.scala
eddyxu: what about make it a `str` value , i.e., `"float64"` for now?  Such values should be dictionary encoded by parquet later.

changhiskhan: sg

eddyxu:
@changhiskhan ready for another round of review.

changhiskhan (approved):

Issue #46, opened 2021-01-24
Should change Label type from str to int


eddyxu:
So that Label can be also applied to the inferred results directly.

eddyxu:
wdyt @changhiskhan

eddyxu:
We can represent today's `Label` to a `Category` or `Class`

changhiskhan:
So Label == StringIndexer(Category) ? I think it can make sense as long as the use case is clear and we can provide clear documentation / conveniences to convert between the two.

eddyxu:
Yes. For example, in batch inference code, the returned values are always numeric values. it is nice to build strong typed output from it.

eddyxu:
@sadhen would like to hear your opinions on this.  In general, we'd like that `rikai` dataset has strong types. That is, while two fields might be the same `Integer` or `String` internally, however, rikai acks that they have a particular meaning in ML setup. 

For example, a `label / class` integer here can help us to build tooling to run aggregation on top of it, because we know there are limited value range of this value, while in the meantime, the value is important for us to analyze the distribution, balance of dataset and etc.

PR #58, opened 2021-01-26
Remove Label definition until we have a better design 

eddyxu:


eddyxu:
This PR moves #46 out of `0.0.1` until we have better understanding.

README.md
```diff
@@ -33,7 +33,7 @@ Roadmap:
 
 ```python
 from pyspark.ml.linalg import DenseMetrix
-from rikai.vision import Image, BBox
+from rikai.types.vision import Image, Box2d
```
changhiskhan, line 36: should users just do from `rikai.types import Image, Box2d` ? I think Box2d is moved to geometry right? Users might find it hard to remember which package to import from?

changhiskhan (approved):

README.md
eddyxu: sure , will change that

PR #61, opened 2021-01-27
Deprecate SparkTestCase and TestCase for pytest fixture

eddyxu:


python/tests/parquet/test_resolver.py
```diff
@@ -12,33 +12,26 @@
...
-import tempfile
-from unittest import TestCase
-
 from rikai.parquet.resolver import Resolver
+from rikai.testing import assert_count_equal
+
 
+def teardown_function(_):
+    Resolver.reset()
+    print("RESET!!")
```
changhiskhan, line 21: super nit: remove print

changhiskhan (approved):
lgtm.

nit: not a release blocker, we could think about using https://pypi.org/project/pytest-spark/ to initialize the test sparkSession. not a blocker tho

python/tests/parquet/test_resolver.py
eddyxu: lol my bad

eddyxu:
Maybe i am too conservative , i hesitate to use project with 500/1000 stars :)

PR #62, opened 2021-01-27
Refactor video sampling API

changhiskhan:
There's some uncertainty in the DataLoader/DataSet API.
Rather than kludging the sampler in, we provide something simple
to start (see VideoData notebook for example). If we need prefetch
capabilities we can abstract that part out of the DataLoader later.
One additional minor refactoring is renaming the show -> display.
After playing in Jupyter notebook, `.display` looked more natural.


python/rikai/spark/functions/geometry.py
```diff
@@ -67,7 +67,7 @@ def box2d_from_top_left(coords) -> Box2d:
     ...     .createDataFrame(coco["annotations"][:5])
     ...     .withColumn("box2d", box2d_from_top_left("bbox"))
     ... )
-    >>> anno_df.show()
+    >>> anno_df.display()
```
eddyxu, line 70: Why this line changed

eddyxu (approved):
Small nits. LGTM overall

python/rikai/spark/functions/geometry.py
changhiskhan: typo from intellij refactor

PR #76, opened 2021-01-29
Remove parameterized package from test

eddyxu:
We can use `pytest.mark.parameterize` now


eddyxu:
https://docs.pytest.org/en/stable/parametrize.html `pytest` has it, remove duplicated deps.

changhiskhan (approved):

PR #77, opened 2021-01-29
Add long_description and other missing fields in setup.py

eddyxu:
Make pypi page with appropriate information.


python/setup.py
```diff
@@ -21,6 +27,10 @@
     version=about["version"],
     license="Apache License, Version 2.0",
     author="Rikai authors",
+    author_email="rikai-dev@eto.ai",
```
changhiskhan, line 30: i forgot did we already make this list?

changhiskhan (approved):

PR #82, opened 2021-01-31
Provide torch Dataset using its IterableDataset interface

eddyxu:
Pytorch has merged a shuffler that works with IterableDataset, we can limit the scope of `rikai.torch.data` to a Dataset instead. It can leave multi-process loading and shuffling to `pytorch` to handle

### DIFF
--- a/python/rikai/parquet/dataset.py
+++ b/python/rikai/parquet/dataset.py
@@ -187,6 +187,13 @@ class Dataset:
         shuffler = RandomShuffler(
             self.shuffler_capacity if self.shuffle else 1, self.seed
         )
+        logger.info(
+            "Start to iterate : rank=%s, world=%s, files=%s",
+            self.rank,
+            self.world_size,
+            self.files,
+        )
+
         group_count = 0
         for filepath in self.files:
             fs, path = FileSystem.from_uri(filepath)
@@ -204,9 +211,11 @@ class Dataset:
                     #   The drawback would be if the world size is much larger
                     #   than the average number of row groups. As a result,
                     #   many of the file open operations would be wasted.
-                    group_count += 1
+                    print("Lets try group count: ", group_count)
                     if group_count % self.world_size != self.rank:
+                        group_count += 1
                         continue
+                    group_count += 1
                     row_group = parquet.read_row_group(
                         group_idx, columns=self.columns
                     )
--- a/python/rikai/parquet/dataset.py
+++ b/python/rikai/parquet/dataset.py
@@ -187,13 +187,6 @@ class Dataset:
         shuffler = RandomShuffler(
             self.shuffler_capacity if self.shuffle else 1, self.seed
         )
-        logger.info(
-            "Start to iterate : rank=%s, world=%s, files=%s",
-            self.rank,
-            self.world_size,
-            self.files,
-        )
-
         group_count = 0
         for filepath in self.files:
             fs, path = FileSystem.from_uri(filepath)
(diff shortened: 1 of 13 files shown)


eddyxu:
When it uses with `torch.utils.data.DataLoader(num_workers>0)` , and the partitions are not _perfectly_ balanced,  data loader hangs.

changhiskhan:
do you want me to review yet? or wait til you finish the work?

eddyxu:
Please wait a bit. I have not figured out what was wrong yet.

eddyxu:
Maybe relevant https://github.com/pytorch/pytorch/issues/15808

eddyxu:
It seems that a few partitions in the dataset returning `_IterableDatasetStopIteration` and then the full dataset stops. 

```
```

python/tests/conftest.py
```diff
@@ -15,6 +15,7 @@
 
 # Third Party
 import pytest
+from torch.utils.data import DataLoader  # Prevent DataLoader hangs
```
changhiskhan, line 18: why does it hang if you don't import it here?

python/tests/parquet/test_shuffler.py
```diff
@@ -13,6 +13,7 @@
 #  limitations under the License.
 
 
+from random import shuffle
```
changhiskhan, line 16: unused?

python/tests/torch/test_data.py
```diff
@@ -18,13 +18,18 @@
...
 from pyspark.sql import Row, SparkSession
+from torch.utils.data import DataLoader as torchDataLoader
 
 # Rikai
 from rikai.numpy import wrap
-from rikai.torch import DataLoader
+from rikai.torch import DataLoader, Dataset
 from rikai.types import Box2d, Image
 
+torch.multiprocessing.set_sharing_strategy("file_system")
```
changhiskhan, line 31: consider putting this in a session scoped fixture or inside pytest_sessionstart ?

python/rikai/torch/data.py
```diff
@@ -13,19 +13,90 @@
...
 
 # Third Party
 import numpy as np
+import torch
+from torch.utils.data import IterableDataset
 
 # Rikai
 from rikai.mixin import ToNumpy
-from rikai.parquet.dataset import Dataset
+from rikai.parquet.dataset import Dataset as pgDataset
```
changhiskhan, line 30: why `pg`?

changhiskhan (approved):

python/rikai/torch/data.py
eddyxu: hmm should be `pq`... for `parquet`

python/tests/conftest.py
eddyxu: I have no good theory, need to talk with the team.
But TLDR after days of trying is that `torch.DataLoader` needs to be imported before` pyspark.sql`

python/tests/parquet/test_shuffler.py
eddyxu: done

python/tests/torch/test_data.py
eddyxu: I am not very familar with the convention for pytest fixture.

Should this be:
```
@pytest.fixture(scope="session"):
def torch_session():
    torch.multiprocessing.set_sharing_strategy("file_system")
```

and let each of the torch test to declare this parameter?

python/rikai/parquet/dataset.py
```diff
@@ -18,6 +18,7 @@
 # Standard Library
 import importlib
 from typing import Any, Dict, List, Optional
+from _pytest.fixtures import yield_fixture
```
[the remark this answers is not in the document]
eddyxu: Removed.

python/tests/conftest.py
eddyxu: Added a `rikai/tests/torch/conftest.py`

python/tests/torch/test_data.py
eddyxu: Added a fixture in `python/tests/torch/conftest.py`

PR #84, opened 2021-02-05
add from_array and from_pil helper functions to Image type

eddyxu:


eddyxu:
Also fixed a few places to use `pathlib.Path` for connivence.

python/rikai/types/vision.py
```diff
@@ -40,10 +49,61 @@ class Image(ToNumpy, Asset, Displayable):
...
+        Parameters
+        ----------
+        img : :py:class:`PIL.Image`
+            An PIL Image instance
+        uri : str or Path
+            The URI to store the image externally.
+        """
+        parsed = urlparse(normalize_uri(uri))
+        if parsed.scheme == "file":
+            img.save(uri)
```
changhiskhan, line 99: I *think* this should also work directly for dbfs

python/rikai/types/vision.py
```diff
@@ -40,10 +49,61 @@ class Image(ToNumpy, Asset, Displayable):
...
+        """Create an image in memory from a :py:class:`PIL.Image`.
+
+        Parameters
+        ----------
+        img : :py:class:`PIL.Image`
+            An PIL Image instance
+        uri : str or Path
+            The URI to store the image externally.
+        """
+        parsed = urlparse(normalize_uri(uri))
```
changhiskhan, line 97: This is something we should probably standardize (there are lots of places whether we have this if this scheme do x else y)

changhiskhan (approved):

python/rikai/types/vision.py
```diff
@@ -40,10 +49,61 @@ class Image(ToNumpy, Asset, Displayable):
...
+        if parsed.scheme == "file":
+            img.save(uri)
+        else:
+            with NamedTemporaryFile() as fobj:
+                img.save(fobj)
+                fobj.flush()
+                copy(fobj.name, uri)
```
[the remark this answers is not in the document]
eddyxu: This works via `copy` (for s3 at least today). 

`rikai.io` supposes to handle the logic of handing different cloud storage backends.

python/rikai/types/vision.py
eddyxu: you meant `PIL` can directly save to `dbfs`?

eddyxu: We could standardize to get a `write fobj / stream` from `uri`, and make them works with these third party libs that works with `fobj`.

python/rikai/types/vision.py
eddyxu: yep, agreed.

PR #86, opened 2021-02-06
A dataset that is compatible with torchvision

eddyxu:


python/rikai/torch/transforms.py
```diff
@@ -12,17 +12,18 @@
...
 
-from rikai.mixin import ToNumpy
+from rikai.mixin import ToNumpy, ToPIL
+
+__all__ = ["RikaiToTensor"]
 
 
-def convert_tensor(row):
-    """Convert row into rikai semantic objects."""
+def convert_tensor(row, use_pil: bool = False):
```
changhiskhan, line 25: nit: the extra boolean parameter is a bit confusing because it took some tracing to figure out where it was being set to True etc.

changhiskhan (approved):
a few nits. no blockers

python/rikai/torch/data.py
```diff
@@ -21,14 +21,12 @@
...
 # Third Party
-import numpy as np
 import torch
 from torch.utils.data import IterableDataset
 
 # Rikai
 import rikai.parquet
-from rikai.mixin import ToNumpy
-from rikai.torch.utils import convert_tensor
+from rikai.torch.transforms import RikaiToTensor, convert_tensor
```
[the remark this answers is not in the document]
eddyxu: This is still used in the Rikai-made `DataLoader`, we can remove it once we start to use torch 1.8's iteratableDataset shuffler

python/rikai/torch/transforms.py
eddyxu: What would be an appropriate way to write parameter here? `use_pil = None` or making it a required parameter?

Issue #91, opened 2021-02-17
Feature request: Support https in Image


chunyang:
Using an HTTPS url with `Image("https://...").to_numpy` gives the following error:

```
Traceback (most recent call last):
  File "repro/repro.py", line 9, in <module>
    "https://upload.wikimedia.org/wikipedia/commons/thumb/a/ad/Commodore_Grace_M._Hopper%2C_USN_%28covered%29.jpg/819px-Commodore_Grace_M._Hopper%2C_USN_%28covered%29.jpg"
  File "/home/chuck.yang/src/repro/venv/lib/python3.7/site-packages/rikai/types/vision.py", line 101, in to_numpy
    with self.to_pil() as pil_img:
  File "/home/chuck.yang/src/repro/venv/lib/python3.7/site-packages/rikai/types/vision.py", line 96, in to_pil
    return PILImage.open(self.open())
  File "/home/chuck.yang/src/repro/venv/lib/python3.7/site-packages/rikai/mixin.py", line 71, in open
    filesystem, path = fs.FileSystem.from_uri(uri)
  File "pyarrow/_fs.pyx", line 347, in pyarrow._fs.FileSystem.from_uri
  File "pyarrow/error.pxi", line 122, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 84, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Unrecognized filesystem type in URI: https://upload.wikimedia.org/wikipedia/commons/thumb/a/ad/Commodore_Grace_M._Hopper%2C_USN_%28covered%29.jpg/819px-Commodore_Grace_M._Hopper%2C_USN_%28covered%29.jpg
```

Minimal reproducible example:
```
from rikai.types.vision import Image


Image(
    "https://upload.wikimedia.org/wikipedia/commons/thumb/a/ad/Commodore_Grace_M._Hopper%2C_USN_%28covered%29.jpg/819px-Commodore_Grace_M._Hopper%2C_USN_%28covered%29.jpg"
).to_numpy()
```

eddyxu:
Closed in #93

PR #93, opened 2021-02-17
Support http / https in Asset URI

eddyxu:
Address #91


python/rikai/mixin.py
```diff
@@ -62,12 +64,4 @@ def __eq__(self, o: object) -> bool:
 
     def open(self, mode="rb") -> BinaryIO:
         """Open the asset and returned as random-accessible file object."""
```
chunyang, line 66: Do you want to provide a note saying that the caller is responsible for closing the returned file object?

chunyang (approved):
LGTM

PR #99, opened 2021-02-19
Support using mul and div to scale Box2d 

eddyxu:
Support `box2d / scale` or `box2d * scale`

### DIFF
--- a/python/rikai/types/geometry.py
+++ b/python/rikai/types/geometry.py
@@ -16,12 +16,14 @@
 """
 
 from __future__ import annotations
-from typing import Union, List
+
+from typing import List, Tuple, Union
+from numbers import Number
 
 import numpy as np
 
 from rikai.mixin import ToNumpy
-from rikai.spark.types.geometry import PointType, Box3dType, Box2dType
+from rikai.spark.types.geometry import Box2dType, Box3dType, PointType
 
 __all__ = ["Point", "Box3d", "Box2d"]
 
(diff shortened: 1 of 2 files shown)


python/rikai/types/geometry.py
```diff
@@ -170,6 +181,59 @@ def __eq__(self, o: Box2d) -> bool:
...
+            return scale, scale
+        assert (
+            type(scale) == tuple and len(scale) == 2
+        ), "scale must be either a number or a 2-element tuple"
+        assert (
+            scale[0] > 0 and scale[1] > 0
+        ), f"scale must be positive, got {scale}"
+        return scale
+
+    def __truediv__(self, scale: Union[int, float, Tuple]) -> Box2d:
```
chunyang, line 199: nit: when both `int` and `float` are acceptable, you can use just `float` if I'm reading this correctly: https://www.python.org/dev/peps/pep-0484/#the-numeric-tower
```suggestion
    def __truediv__(self, scale: Union[float, Tuple[float, float]]) -> Box2d:
```

python/rikai/types/geometry.py
```diff
@@ -170,6 +181,59 @@ def __eq__(self, o: Box2d) -> bool:
             self.to_numpy(), o.to_numpy()
         )
 
+    @staticmethod
+    def _verified_scale(
+        scale: Union[int, float, Tuple]
+    ) -> Tuple[Number, Number]:
```
chunyang, line 187: nit: probably `float` instead of `Number`? Since `Number` can be fractional or complex.

chunyang (approved):

python/rikai/types/geometry.py
eddyxu: Changed to `numbers.Real` , it needs to be able to handle `numpy` types

PR #100, opened 2021-02-19
Support display Image with https link in Juptyer

eddyxu:


eddyxu:
Screen Shot 2021-02-18 at 7 19 22 PM

python/rikai/types/vision.py
```diff
@@ -119,7 +116,8 @@ def display(self, **kwargs):
         """
         from IPython.display import Image
 
-        return Image(self.uri, **kwargs)
```
chunyang, line 122: Curious why many libraries don't support loading from HTTPs out of the box (Jupyter, TensorFlow, others?). Jupyter appears to support http but not https?

chunyang (approved):
:shipit:

python/rikai/types/vision.py
eddyxu: Usually a good design should design against to a "file-object" like interface, i'd say.

PR #103, opened 2021-02-20
[SQL] Add ML Model in SQL Catalog

eddyxu:
Define traits for `Model`, `Registry` and `(ML)Catalog`


src/main/antlr4/org/apache/spark/sql/ml/parser/RikaiExtSqlBase.g4
```diff
@@ -22,7 +22,7 @@ singleStatement
 statement
     : CREATE (OR REPLACE)? MODEL model=qualifiedName
       (OPTIONS optionList)?
-      (AS table=qualifiedName | USING path=STRING)	# createModel
+      (USING path=STRING)	                        # createModel
```
eddyxu, line 25: Start from supporting URIs

src/main/scala/ai/eto/rikai/sql/model/Model.scala
```diff
@@ -0,0 +1,66 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ai.eto.rikai.sql.model
+
+import org.apache.spark.sql.catalyst.expressions.Expression
```
changhiskhan, line 19: how tightly do we want to couple this with Spark?
e.g., would we ever want to integrate the ML SQL into say Presto or something? (assuming we're able to install and execute the models somehow).

changhiskhan (commented):
do we want to try implementing catalogs backed by S3 / MLFlow Tracking to help improve the API?

src/main/scala/ai/eto/rikai/sql/model/Model.scala
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+  * A Machine Learning Model in Rikai Catalog.
+  */
+trait Model {
+
+  /** Model Name */
+  val name: String
+
+  /** Model URI in the registry */
```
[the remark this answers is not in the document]
eddyxu: Catalog is basically a list of "imported models" within the SparkSession.  And a registry is an existing external service which can refer a model by a URL.   For example, `MLFlow` is one model registry, and we can make filesystem / cloud storage another one.

`Catalog` is for SQL parser to resolve symbols and find correct logic to execute IN THE SQL engine. It is used for when you run `SELECT ML_PREDICT(model_foo, col1, col2, col3)`, we can find what `model_foo` is and how to execute it.

src/main/scala/ai/eto/rikai/sql/model/Model.scala
```diff
@@ -0,0 +1,66 @@
...
+    */
+  def expr(arguments: Seq[Expression]): Expression
+
+}
+
+object Model {
+
+  /**
+    */
+  val namePattern = """[a-zA-Z]\w{0,255}""".r
```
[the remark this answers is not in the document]
eddyxu: `\w` is `[a-zA-Z_0-9]`

https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html

src/main/scala/ai/eto/rikai/sql/model/Model.scala
eddyxu: In a foresee future, I'd say it will be very tight (thus Spark SQL only), given how much integrations we need to do to provide the "Native feel" for any SQL engine.

eddyxu:
> Where does registry and catalog come into play?

Replied inline above

> Can we have multiple catalogs in one spark sessions? If so does SHOW MODELS show all of them?

For one spark session, we should have 1 `Catalog`, but it can support mulitple Registry? Registry is similar to what a cloud connector is in the traditional Spark term. Given a `uri`, a `Registry` returns you a model , which can generate logical plan later.   So we can support different `Registry` using different uri schema , ie., `mlflow://` for `MLFlowRegistry`.

> I'm not sure drop/create makes sense here. I was thinking that the SQL syntax would just expose a read-only API to the underlying registry/catalog but the actual models would be added by some training pipeline separately.

I think the main question here is that whether `Model Catalog` is a proxy of an external ML registry, or `Catalog` is a separately managed entity to a `Model Registry`

eddyxu:
BigQuery ML's [`CREATE MODEL`](https://cloud.google.com/bigquery-ml/docs/reference/standard-sql/bigqueryml-syntax-create). and Redshift [one](https://docs.aws.amazon.com/redshift/latest/dg/simple-training.html)

@changhiskhan

eddyxu:
SQL-ML architecture

eddyxu:
take another look @changhiskhan ?  Split `model => spark expression` conversion into a separate trait `ai.eto.rikai.sql.spark.SparkRunnable`

src/main/scala/ai/eto/rikai/sql/package.scala
```diff
@@ -0,0 +1,39 @@
...
+  *   SHOW MODELS
+  *
+  *   # Describe the details of a model.
+  *   (DESC | DESCRIBE) MODEL model_name
+  *
+  *   # Drop a Model
+  *   DROP MODEL model_name
+  * }}}
+  */
+package object sql {}
```
bobingm, line 39: what is this file for?

eddyxu: This is used for generating package level document.

changhiskhan (approved):
lgtm. we can evolve this when we add mlflow and s3 integrations

PR #107, opened 2021-02-23
Add options parsing and show them in show models

bobingm:
## Summary

This PR is to 
1. Add `optionList` parsing logic
2. Show option list in `SHOW MODELS` query

### DIFF
--- a/src/main/scala/ai/eto/rikai/sql/execution/ShowModelsCommand.scala
+++ b/src/main/scala/ai/eto/rikai/sql/execution/ShowModelsCommand.scala
@@ -6,10 +6,16 @@ import ai.eto.rikai.sql.catalog.{MLCatalog, Model}
 import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference}
 import org.apache.spark.sql.types.StringType
 
+import org.json4s._
+import org.json4s.jackson.Serialization
+import org.json4s.jackson.Serialization.write
+
+
 case class ShowModelsCommand() extends RunnableCommand {
     override val output: Seq[Attribute] = Seq(
         AttributeReference("Model Name", StringType, nullable = false)(),
-        AttributeReference("Model Path", StringType, nullable = false)()
+        AttributeReference("Model Path", StringType, nullable = false)(),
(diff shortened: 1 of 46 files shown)


src/main/scala/ai/eto/rikai/sql/execution/ShowModelsCommand.scala
```diff
@@ -6,10 +6,20 @@ import ai.eto.rikai.sql.catalog.{MLCatalog, Model}
...
+
+
 case class ShowModelsCommand() extends RunnableCommand {
+
+    implicit val formats = Serialization.formats(NoTypeHints)
+
     override val output: Seq[Attribute] = Seq(
         AttributeReference("Model Name", StringType, nullable = false)(),
-        AttributeReference("Model Path", StringType, nullable = false)()
+        AttributeReference("Model Path", StringType, nullable = false)(),
```
eddyxu, line 21: We can just use `Name, Path, Options`  fo the name?

Are `path` and `options` nullable?

eddyxu: In the `master` branch, we introduced a [`SparkRunnable`](https://github.com/eto-ai/rikai/blob/master/src/main/scala/ai/eto/rikai/sql/spark/SparkRunnable.scala), we might want to only display `SparkRunnable`s in ```SHOW MODELS```?

src/main/scala/ai/eto/rikai/sql/parser/RikaiExtAstBuilder.scala
```diff
@@ -42,6 +42,16 @@ class RikaiExtAstBuilder extends RikaiSqlBaseBaseVisitor[AnyRef] {
       }
     }
 
+  protected def parseOptionList(ctx: OptionListContext): Map[String, String] = {
+    if (ctx != null) {
```
eddyxu, line 46: can ctx be null?

bobingm: Yes, here we use `OptionListContext`, if create model query doesn't have `OPTIONS`, this will be a null.

PR #112, opened 2021-02-27
[SQL] Add the capability to JIT python code for a Model 

eddyxu:
Re-use `py4j` in `SparkSession` to register callback in `SparkSession`'s  Python interpreter, so that later we can generate JIT code for models in python.


eddyxu:
Run `pytest python/tests/spark/test_code_gen.py -s`

 to demonstrate the round trips between JVM and python

python/rikai/spark/sql/session.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.started:
+            self.stop()
+
+    def start(self):
+        self.spark.sparkContext._gateway.start_callback_server()
+        logger.info("Spark callback server started")
+
+        self.callback_service.register()
```
eddyxu, line 45: The design choice would be either 

- start a rpc/rest server locally, pass the service uri via SparkSession. 
- re-use `py4j` in Spark and open callback server.

eddyxu:
The `py4j` callback server is daemonized now. SparkSession can exit normally

python/rikai/spark/sql/callback_service.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+    jvm.start_callback_server(callback_server_parameters=params)
+    logger.info("Spark callback server started")
+
+    cb = CallbackService(spark)
+    cb.register()
+    logger.info("Rikai Python callback service registered")
+
+
+class CallbackService(object):
```
changhiskhan, line 39: nit: i don't think it's required anymore to inherit from `object`

python/rikai/spark/sql/callback_service.py
```diff
@@ -0,0 +1,82 @@
...
+            The model to generate python code.
+        temporary : bool
+            Set true of the generated code will only be used once. Temporary
+            code generation will be used when we use `ML_PREDICT` directly
+            with a model URI.
+        """
+        opt = model.javaOptions()
+        options = {key: opt[key] for key in opt}
+        py_class = model.pyClass()
+        print(model.toString(), options, py_class, temporary)
```
changhiskhan, line 69: logger.debug?

python/rikai/spark/sql/callback_service.py
```diff
@@ -0,0 +1,82 @@
...
+    Internal use only
+    """
+
+    def __init__(self, spark: SparkSession):
+        self.spark = spark
+
+    def __repr__(self):
+        return "PythonCbService"
+
+    def codegen(self, model, temporary: bool):
```
changhiskhan, line 54: where/when does this generated code get used?

changhiskhan (approved):
So based on our discussion the problem we're trying to solve is that the model lives in python but the main Spark SQL execution with ML_PREDICT lives in Scala and so this is the rpc to cross the language barrier?

python/rikai/spark/sql/callback_service.py
eddyxu: When a model first time introduced in the system, for example:

```
CREATE MODEL foo USING 'uri://to/path'
```

Or 

```
SELECT ML_PREDICT("uri://to/path", col) FROM t
```

In both cases, if the model needs to run in python,   `Model.pyClass.codegen()` will generate udfs, prepare enviornment and etc , and register the udfs into the main SparkSession in pyspark. So for the Spark SQL AST, it can replace the `ML_PREDICT` expression to a UDF that generated:

```
SELECT ML_PREDICT("uri://to/path/model", col) FROM t
=>
SELECT model_udf_tmp12345(col) FROM t
```

Each  `Model / Registry` will implement its own way to run the model, for example, we can implement a [vanilla pytorch inference](https://docs.databricks.com/_static/notebooks/deep-learning/pytorch-images.html) using `@pandas_udf` and `torch.DataLoader`,  or a UDF that sends http / gprc requests to an online inference server.

python/rikai/spark/sql/callback_service.py
eddyxu: done

python/rikai/spark/sql/callback_service.py
eddyxu: fixed

eddyxu:
Yes this PR establishes a communication from Spark's JVM (Spark SQL parser) to PySpark python interpreter.

python/rikai/spark/sql/callback_service.py
eddyxu: fixed

python/rikai/spark/sql/callback_service.py
eddyxu: done

PR #116, opened 2021-03-06
[annotations] Proposed spec for annotations API

changhiskhan:
Notebook with an outline for annotation spec.


eddyxu:
What is the proposed workflow? Users first save the rikai format in storage,  load them back, and provide the declaration attach to those UDTs?

changhiskhan:
moved to google docs.

PR #118, opened 2021-03-08
Raise exception if model already exist

bobingm:
## Summary

Raise `ModelAlreadyExistException` when creating model.


eddyxu (approved):
LGTM

PR #119, opened 2021-03-11
[SQL] A YAML file-based model

eddyxu:
This PR lays the skeleton of using a YAML-based spec file to import models.


python/rikai/spark/sql/codegen/fs.py
```diff
@@ -0,0 +1,89 @@
...
+from rikai.io import open_uri
+from rikai.logging import logger
+from rikai.spark.sql.codegen.exceptions import ModelSpecFormatError
+from rikai.spark.sql.codegen.runners import pytorch_runner
+
+
+def codegen_from_yaml(
+    spark: SparkSession, yaml_uri: str, name: Optional[str] = None
+):
+    # TODO(lei): find a schema validation for spec?
```
eddyxu, line 31: Suggestion for a library for YAML scheme check? @changhiskhan

python/tests/spark/test_code_gen.py
```diff
@@ -11,15 +11,58 @@
...
 
+def test_model_codegen_registered(spark: SparkSession):
     spark.sql(
         """CREATE MODEL foo OPTIONS (foo="str",bar=True,max_score=1.23)
          USING 'test://model/a/b/c'"""
     ).count()
+
+
+def test_yaml_spec(spark: SparkSession, tmp_path: Path):
+    spec_yaml = """
```
eddyxu, line 34: @chunyang would like to see your inputs on the spec file

python/rikai/spark/sql/codegen/fs.py
```diff
@@ -0,0 +1,89 @@
...
+            f"Miss model.uri from YAML spec: {yaml_uri}"
+        ) from e
+    if model_uri is None:
+        raise ModelSpecFormatError(
+            f"Miss model.uri from YAML spec: {yaml_uri}"
+        )
+
+    try:
+        schema_str = spec["schema"]
+        schema = StringType()  # parse_schema(schema_str)
```
eddyxu, line 59: This makes me think we should implement #115 in python.

python/tests/spark/test_code_gen.py
```diff
@@ -11,15 +11,60 @@
...
+version: 1.0
+name: resnet
+model:
+  uri: resnet.pth
+  flavor: pytorch
+schema: struct<box:box2d, score:float, class:int>
+transforms:
+  pre: demoproject.yolo.transform
+  post: demoproject.yolo.postprocess
```
chunyang, line 43: Some thoughts:
* Should we also make the input schema explicit? There's actually several boundaries where one might wish to know the shape/schema of the data: 1) before pre-processing, 2) after pre-processing, 3) before post-processing, 4) after post-processing. If we know 1 and 2 then maybe we can automatically build a pre-processing pipeline (e.g., if it only involves resizing and batching images) but maybe that's wishful thinking (in practice the steps might not be so trivial).
* Should we provide the ability to construct the declaratively in YAML? Something like
  ```
  transforms:
    pre:
      - resize
          width: 600
          height: 600
          mode: fill
      - batch:
          size: 4
  ```
  This way the model creator doesn't have to write imperative preprocessing code.

eddyxu:
This is done in #131 and #135

PR #124, opened 2021-03-14
[SQL] Fix models being not resolvable as attribute

eddyxu:
The model that registered with Catalog was not resolvable from later `ML_PREDICT` function because it mistakenly not detect Optional[Model] as return from catalog.

### DIFF
--- a/src/test/scala/ai/eto/rikai/sql/spark/parser/RikaiSparkSQLParserTest.scala
+++ b/src/test/scala/ai/eto/rikai/sql/spark/parser/RikaiSparkSQLParserTest.scala
@@ -39,10 +39,12 @@ class RikaiSparkSQLParserTest extends AnyFunSuite with SparkTestSession {
   }
 
   test("Test parse ML_PREDICT with catalog") {
-    spark.sql("CREATE MODEL foo USING 'test://host/foo'").show()
+    spark.udf.register("bar", (s: Int) => s + 2)
+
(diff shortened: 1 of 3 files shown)


chunyang (approved):
LGTM

src/test/scala/ai/eto/rikai/sql/spark/parser/RikaiSparkSQLParserTest.scala
```diff
@@ -39,10 +39,12 @@ class RikaiSparkSQLParserTest extends AnyFunSuite with SparkTestSession {
...
 
   test("Test parse ML_PREDICT with catalog") {
-    spark.sql("CREATE MODEL foo USING 'test://host/foo'").show()
+    spark.udf.register("bar", (s: Int) => s + 2)
+
+    spark.sql("CREATE MODEL bar USING 'test://host/bar'").show()
 
     val scores =
-      spark.sql("SELECT id, ML_PREDICT('test://host/foo', id) AS score FROM df")
+      spark.sql("SELECT id, ML_PREDICT(bar, id) AS score FROM df")
```
changhiskhan, line 47: why wouldn't `bar` be quoted here?

eddyxu: it was a mistake by copying the other tests and forgot to change to a reference to an existing model instead of a temporary model referred by string.

PR #126, opened 2021-03-16
[SQL] Implement FileSystemRegistry

eddyxu:
Productionalize YAML-based file model implementation from #119

### DIFF
--- a/python/rikai/spark/sql/callback_service.py
+++ b/python/rikai/spark/sql/callback_service.py
@@ -66,18 +66,21 @@ class CallbackService:
             The model URI.
         name : str
             Mode name
-        options : JavaMap
+        options : dict[str, str]
             Options passed to the model.
         """
-        try:
-            registry = self.registry_map[registry_class]
-        except KeyError:
+        if registry_class not in self.registry_map:
             from rikai.internal.reflection import find_class
+            from rikai.spark.sql.codegen.base import Registry
 
             cls = find_class(registry_class)
+            if not issubclass(cls, Registry):
+                raise ValueError(
+                    f"Class '{registry_class}' is not a Registry'"
+                )
             self.registry_map[registry_class] = cls(self.spark)
-            registry = self.registry_map[registry_class]
 
+        registry = self.registry_map[registry_class]
         options = {key: options[key] for key in options}
         return registry.resolve(uri, name, options)
 
--- a/python/rikai/spark/sql/callback_service.py
+++ b/python/rikai/spark/sql/callback_service.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 
 from py4j.java_gateway import CallbackServerParameters
+from py4j.java_collections import JavaMap
 from pyspark.sql import SparkSession
 
 from rikai.logging import logger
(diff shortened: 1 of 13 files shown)


eddyxu:
Will work on a pandas UDF generation in the next PR

python/rikai/spark/sql/callback_service.py
```diff
@@ -47,29 +48,38 @@ class CallbackService:
...
-        options = {key: opt[key] for key in opt}
-        py_class = model.pyClass()
-        logger.debug(
-            f"Code generation for model={model.toString()}, options={options}"
-            f", pyClass=${py_class}, temporary=${temporary}"
-        )
+        try:
+            registry = self.registry_map[registry_class]
+        except KeyError:
+            from rikai.internal.reflection import find_class
```
changhiskhan, line 75: nit: this isn't an unexpected case right? consider using `if registry_class in self.registry_map` instead?

python/rikai/spark/sql/callback_service.py
```diff
@@ -47,29 +48,38 @@ class CallbackService:
...
+
+            cls = find_class(registry_class)
+            self.registry_map[registry_class] = cls(self.spark)
```
changhiskhan, line 78: would it make sense to create an ABC Registry with `resolve` method?

changhiskhan: and we can also do an `issubclass` check here?

changhiskhan: i know i might be overly paranoid but these are pretty complicated internals so might be worth being more defensive / help debugging.

src/main/scala/ai/eto/rikai/sql/model/fs/FileSystemRegistry.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.logging.log4j.scala.Logging
+
+/**
+  * FileSystem-based Model [[Registry]].
+  */
+class FileSystemRegistry(val conf: Map[String, String])
+    extends Registry
+    with Logging {
+
+  private val pyClass: String = "rikai.spark.sql.codegen.fs.Registry"
```
changhiskhan, line 30: All of these raw string references are going to suck if we have to refactor/move packages. Not sure what's a better alternative tho.

changhiskhan (approved):
Some suggestions

python/rikai/spark/sql/codegen/fs.py
```diff
@@ -0,0 +1,42 @@
...
+    """FileSystem-based Model Registry"""
+
+    def __init__(self, spark: SparkSession):
+        self.spark = spark
+        self.jvm = spark.sparkContext._jvm
+
+    def __repr__(self):
+        return f"FileSystemRegistry"
+
+    def resolve(self, uri: str, name: str, options):
```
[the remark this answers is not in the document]
eddyxu: it is a little bit tricky that it returns a java class (py4j wrapped). Because this method is basically directly called from JVM. We can probally add one layer of abstraction to return here, and convert it to JVM class right before `py4j` returns it tho. What do you think?

python/rikai/spark/sql/codegen/fs.py
```diff
@@ -0,0 +1,42 @@
...
+
+from rikai.logging import logger
+
+
+class Registry:
+    """FileSystem-based Model Registry"""
+
+    def __init__(self, spark: SparkSession):
+        self.spark = spark
+        self.jvm = spark.sparkContext._jvm
```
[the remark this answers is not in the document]
eddyxu: You meant

```
@property
def jvm(self):
   return self._jvm
```
?
This is only used internally.  If we'd like to use `_` prefix to private member, I can change `self.spark` and `self.jvm` to this form as well.

python/rikai/spark/sql/callback_service.py
eddyxu: Right, makes sense.

python/rikai/spark/sql/callback_service.py
eddyxu: Make total sense.  Will fix them

src/main/scala/ai/eto/rikai/sql/model/fs/FileSystemRegistry.scala
eddyxu: Yea, i agree. But this is also give the flexibility that if a third party wanted to add anything (i.e., extensions) to a released `rikai`, it is relatively easy.

PR #127, opened 2021-03-16
video_to_images using Segment

smellslikeml:
The Kinetics dataset links youtube_ids to video segments labeled by activity:
image

Adding the optional segment argument to `video_to_images` helps deal with variable length video segments. 
image


changhiskhan (approved):
How are you converting from the time_start / time_end to start/end frame numbers? That might be another useful utility to add.

smellslikeml:
> How are you converting from the time_start / time_end to start/end frame numbers? That might be another useful utility to add.

ffmpeg models this with `-start_number` versus `-ss` for start time in seconds.

PR #129, opened 2021-03-17
video to spectrogram udf w ffmpeg

smellslikeml:
[ffmpeg-python](https://github.com/kkroening/ffmpeg-python) offers another way to convert video into arrays, while supporting other common video processing workloads which can be implemented as an ffmpeg signal graph.

This example renders a spectrogram
image

We can cast the result as an Image:
image

This PR implements the udf for rendering audio in ML-friendly formats like Image:
image

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -30,6 +30,7 @@ jobs:
       working-directory: python
       run: |
         sudo apt-get -y -qq install libsnappy-dev
+        sudo apt-get -y install ffmpeg
         python -m pip install -q wheel pylint black flake8
         python -m pip install -e .[all]
     - name: Python black and flake8
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -29,8 +29,7 @@ jobs:
     - name: Pip install
       working-directory: python
       run: |
-        sudo apt-get -y -qq install libsnappy-dev
-        sudo apt-get -y install ffmpeg
+        sudo apt-get -y -qq install libsnappy-dev ffmpeg
         python -m pip install -q wheel pylint black flake8
         python -m pip install -e .[all]
     - name: Python black and flake8
--- a/python/rikai/spark/functions/vision.py
+++ b/python/rikai/spark/functions/vision.py
@@ -16,7 +16,6 @@
 """
 
 # Third Party
-import ffmpeg
 import numpy as np
 from pyspark.sql.functions import udf
 from pyspark.sql.types import ArrayType
@@ -103,6 +102,7 @@ def video_to_images(
     video, sample_rate: int = 1, start_frame: int = 0, max_samples: int = 15000
 ) -> list:
     """Extract video frames into a list of images.
+
     Parameters
     ----------
     video : Video
@@ -113,6 +113,7 @@ def video_to_images(
         Start from a specific frame
     max_samples : Int
         Yield at most this many frames (-1 means no max)
+
     Return
     ------
     List
@@ -144,16 +145,22 @@ def video_to_images(
 
 
 @udf(returnType=ImageType())
-def spectrogram_image(video) -> Image:
-    """Applied ffmpeg filter to generate spectrogram
+def spectrogram_image(video, size: Int = 224) -> Image:
+    import ffmpeg
+
+    """Applied ffmpeg filter to generate spectrogram.
+
     Parameters
     ----------
     video : Video
         A video object, either YouTubeVideo or VideoStream.
+    size : Int
+        Sets resolution of frequency, time spectrogram image.
+
     Return
     ------
     Image
-        Return an Image of the audio spectrogram
+        Return an Image of the audio spectrogram.
     """
     assert isinstance(video, YouTubeVideo) or isinstance(
         video, VideoStream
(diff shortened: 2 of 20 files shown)


python/rikai/spark/functions/vision.py
```diff
@@ -16,6 +16,8 @@
 """
 
 # Third Party
+import ffmpeg
```
eddyxu, line 19: Hi, is `ffmpeg` a required dependency of rikai or an optional one？
Should we do lazy import of `ffmpeg` if this is not a required dependency, so that it does not leak to non-video cases.

python/rikai/spark/functions/vision.py
```diff
@@ -133,3 +141,37 @@ def video_to_images(
         )
         for idx, img in enumerate(video_iterator)
     ]
+
+
+@udf(returnType=ImageType())
+def spectrogram_image(video) -> Image:
+    """Applied ffmpeg filter to generate spectrogram
```
eddyxu, line 148: Lets add blank lines between each sections.

python/rikai/spark/functions/vision.py
```diff
@@ -133,3 +141,37 @@ def video_to_images(
...
+
+    base_path = video.vid if isinstance(video, YouTubeVideo) else video.uri
+    video_uri = (
+        video.get_stream().uri
+        if isinstance(video, YouTubeVideo)
+        else video.uri
+    )
+    output, _ = (
+        ffmpeg.input(video_uri)
+        .filter("showspectrumpic", "224x224", legend=0)
```
eddyxu, line 170: Are these filters configurable later?

python/rikai/spark/functions/vision.py
smellslikeml: That sounds good.

python/rikai/spark/functions/vision.py
smellslikeml: Added a parameter to set default size.
The time dimension of the spectrogram will scale with video duration and can be resized or trimmed.

.github/workflows/python.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       working-directory: python
       run: |
         sudo apt-get -y -qq install libsnappy-dev
+        sudo apt-get -y install ffmpeg
```
eddyxu, line 33: should we merge this line with the above line?

python/rikai/spark/functions/vision.py
```diff
@@ -133,3 +142,43 @@ def video_to_images(
         )
         for idx, img in enumerate(video_iterator)
     ]
+
+
+@udf(returnType=ImageType())
+def spectrogram_image(video, size: int = 224) -> Image:
+    import ffmpeg
```
eddyxu, line 149: this import should be written below docstring.

eddyxu (approved):
Small nits,  LGTM overall

python/rikai/spark/functions/vision.py
```diff
@@ -133,3 +142,43 @@ def video_to_images(
...
+        A video object, either YouTubeVideo or VideoStream.
+    size : Int
+        Sets resolution of frequency, time spectrogram image.
+
+    Return
+    ------
+    Image
+        Return an Image of the audio spectrogram.
+    """
+    assert isinstance(video, YouTubeVideo) or isinstance(
```
changhiskhan, line 165: `isinstance(video, (YouTubeVideo, VideoStream))` should work

python/rikai/spark/functions/vision.py
```diff
@@ -133,3 +142,43 @@ def video_to_images(
         )
         for idx, img in enumerate(video_iterator)
     ]
+
+
+@udf(returnType=ImageType())
+def spectrogram_image(video, size: int = 224) -> Image:
+    import ffmpeg
+
+    """Applied ffmpeg filter to generate spectrogram.
```
changhiskhan, line 151: add a link explaining what a audio spectrogram is used for?

python/setup.py
```diff
@@ -18,7 +18,7 @@
 jupyter = ["matplotlib", "jupyterlab"]
 aws = ["boto"]
 docs = ["sphinx"]
-youtube = ["pafy", "youtube_dl"]
+youtube = ["pafy", "youtube_dl", "ffmpeg-python"]
```
changhiskhan, line 21: this shouldn't go in `youtube` extras. 
how big is the ffmpeg package?
@eddyxu should we think about ffmpeg as an essential library like say opencv? or an optional library?

changhiskhan (commented):
nice work. what's our use case for extracting spectrogram from videos?

python/rikai/spark/functions/vision.py
smellslikeml: This [KDNuggets post](https://www.kdnuggets.com/2020/02/audio-data-analysis-deep-learning-python-part-1.html) has some nice exposition on building models for audio classification. Practitioners often use the spectrogram to featurize an audio sample. Then pairing with CNNs, the audio classification task can be recast as an image classification one.

.github/workflows/python.yml
smellslikeml: 👍

python/rikai/spark/functions/vision.py
smellslikeml: sounds good

python/rikai/spark/functions/vision.py
smellslikeml: Added a segment parameter to help configure time range.

python/rikai/spark/functions/vision.py
smellslikeml: got it

python/rikai/spark/functions/vision.py
changhiskhan: @smellslikeml would it make sense to generate a content audio embedding using the spectrogram and see if that gives us a unique signal?

smellslikeml: A spectrogram image is a simple way to represent audio. People often use other audio/signal processing libraries, specialized filter banks, and even [learnable ones](https://ai.googleblog.com/2021/03/leaf-learnable-frontend-for-audio.html).  Fourier transforms have an advantage in allowing us to recover the original audio by [inversion](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.istft.html).

PR #131, opened 2021-03-18
[SQL] Parse model spec in YAML format

eddyxu:
* Provide a schema validation for Model Spec YAML format.
* a `ModelSpec` object offers easy access to model spec data.


python/rikai/spark/sql/callback_service.py
```diff
@@ -34,7 +34,6 @@ def init_cb_service(spark: SparkSession):
 
     cb = CallbackService(spark)
     cb.register()
-    logger.info("Rikai Python callback service registered")
```
eddyxu, line 37: This log has been printed in `cb.register()` as well.

python/rikai/spark/sql/codegen/fs.py
```diff
@@ -13,12 +13,121 @@
...
+    with open_uri(uri) as fobj:
+        spec = ModelSpec(fobj)
+
+    if spec.version != 1.0:
+        raise SpecError(
+            f"Only spec version 1.0 is supported, got {spec.version}"
+        )
+
+    if spec.flavor == "pytorch":
+        pass
```
changhiskhan, line 124: what's this?

python/rikai/spark/sql/codegen/fs.py
```diff
@@ -13,12 +13,121 @@
...
+# YAML-Spec SCHEMA
+SPEC_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "version": {
+            "type": "number",
+            "description": "Model SPEC format version",
+        },
+        "name": {"type": "string", "description": "Model name"},
+        "schema": {"type": "string"},
```
changhiskhan, line 39: i forgot what this schema is?

changhiskhan (approved):

python/rikai/spark/sql/codegen/fs.py
eddyxu: For every flavor, it will build a UDF to run inference differently.

python/rikai/spark/sql/codegen/fs.py
eddyxu: This is used to build the return type of a UDF (pandas_udf) in this case. 

The POC code can be seen here: https://github.com/eto-ai/rikai/pull/119/files#diff-7fccc4b1dbc18bbb3a690fc62930ecddeb86c0f17dc274555a6a4affa61f7e8dR65

python/rikai/spark/sql/codegen/fs.py
eddyxu: https://github.com/eto-ai/rikai/pull/119/files#diff-7fccc4b1dbc18bbb3a690fc62930ecddeb86c0f17dc274555a6a4affa61f7e8dR65

Issue #130, opened 2021-03-18
Use tmp_path in video tests


eddyxu:
Lets use `tmp_path` fixture in pytest for tests where temporary files are generated. 

Right now, the tests generated files under the working directories. 

![Screen Shot 2021-03-17 at 11 40 45 PM](https://user-images.githubusercontent.com/17097/111583855-421eb280-877a-11eb-931d-0c86d17f8f63.png)

PR #133, opened 2021-03-18
add tmp dir for YouTubeVideo

smellslikeml:
Addressing #130, added tmp dir to base path for ```Image``` saved from ```YouTubeVideo```

### DIFF
--- a/python/rikai/spark/functions/vision.py
+++ b/python/rikai/spark/functions/vision.py
@@ -14,10 +14,10 @@
 
 """Vision related Spark UDFs.
 """
+from typing import Union
 
 # Third Party
 import os
-import tempfile
 import numpy as np
 from pyspark.sql.functions import udf
 from pyspark.sql.types import ArrayType
@@ -106,11 +106,11 @@ def numpy_to_image(array: ndarray, uri: str) -> Image:
 
 @udf(returnType=ArrayType(ImageType()))
 def video_to_images(
-    video,
+    video: Union[VideoStream, YouTubeVideo],
+    output_uri: str,
     segment: Segment = Segment(0, -1),
     sample_rate: int = 1,
     max_samples: int = 15000,
-    output_dir: str = "",
     quality: str = "worst",
 ) -> list:
     """Extract video frames into a list of images.
@@ -119,15 +119,14 @@ def video_to_images(
     ----------
     video : Video
         An video object, either YouTubeVideo or VideoStream.
-    segment: Segment
+    output_uri: str
+        Frames will be written as <output_uri>/<fno>.jpg
+    segment: Segment, default Segment(0, -1)
         A Segment object, localizing video in time to (start_fno, end_fno)
-    sample_rate : int
-        The sampling rate in number of frames
-    max_samples : int
-        Yield at most this many frames (-1 means no max)
-    output_dir: str
-        The output directory where images from YouTubeVideo or
-        VideoStream will be written.
+    sample_rate : int, default 1
+        Keep 1 out of every sample_rate frames.
+    max_samples : int, default 15000
+        Return at most this many frames (-1 means no max)
     quality: str, default 'worst'
         Either 'worst' (lowest bitrate) or 'best' (highest bitrate)
         See: https://pythonhosted.org/Pafy/index.html#Pafy.Pafy.getbest
@@ -142,16 +141,11 @@ def video_to_images(
     ), "Input type must be YouTubeVideo or VideoStream"
     assert isinstance(segment, Segment), "Second input type must be Segment"
 
-    base_path = os.path.join(
-        output_dir, os.path.splitext(os.path.basename(video.uri))[0]
-    )
-
     start_frame = segment.start_fno
     if segment.end_fno > 0:
         max_samples = min((segment.end_fno - start_frame), max_samples)
 
     if isinstance(video, YouTubeVideo):
-        base_path = os.path.join(output_dir, video.vid)
         video_iterator = SingleFrameSampler(
             video.get_stream(quality=quality),
             sample_rate,
@@ -166,7 +160,9 @@ def video_to_images(
     return [
         Image.from_array(
             img,
-            "{}_{}.jpg".format(base_path, (start_frame + idx) * sample_rate),
+            os.path.join(
+                output_uri, "{}.jpg".format((start_frame + idx) * sample_rate)
+            ),
         )
         for idx, img in enumerate(video_iterator)
     ]
@@ -174,7 +170,8 @@ def video_to_images(
 
 @udf(returnType=ImageType())
 def spectrogram_image(
-    video,
+    video: Union[VideoStream, YouTubeVideo],
+    output_uri: str,
     segment: Segment = Segment(0, -1),
     size: int = 224,
     max_samples: int = 15000,
@@ -183,8 +180,10 @@ def spectrogram_image(
 
     Parameters
     ----------
-    video : Video
-        A video object, either YouTubeVideo or VideoStream.
+    video : VideoStream or YouTubeVideo
+        A video object whose audio track will be converted to spectrogram
+    output_uri: str
+        The uri to which the spectrogram image will be written to
     segment: Segment
             A Segment object, localizing video in time to (start_fno, end_fno)
     max_samples : Int
@@ -197,14 +196,19 @@ def spectrogram_image(
     Image
         Return an Image of the audio spectrogram.
     """
-    import ffmpeg
-
+    try:
+        import ffmpeg
+    except ImportError:
+        raise ValueError(
+            "Couldn't import ffmpeg. Please make sure to "
+            "`pip install ffmpeg-python` explicitly or install "
+            "the correct extras like `pip install rikai[all]`"
+        )
     assert isinstance(
         video, (YouTubeVideo, VideoStream)
     ), "Input type must be YouTubeVideo or VideoStream"
     assert isinstance(segment, Segment), "Second input type must be Segment"
 
-    base_path = video.vid if isinstance(video, YouTubeVideo) else video.uri
     start_frame = segment.start_fno
     if segment.end_fno > 0:
         max_samples = min((segment.end_fno - start_frame), max_samples)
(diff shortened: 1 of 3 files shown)


python/rikai/spark/functions/vision.py
```diff
@@ -117,7 +119,7 @@ def video_to_images(
     base_path = video.uri
 
     if isinstance(video, YouTubeVideo):
-        base_path = video.vid
+        base_path = os.path.join(tempfile.gettempdir(), video.vid)
```
eddyxu, line 122: Thanks for addressing #130 .  This as in production code, will it make all data writting to the spark worker's local file system?

changhiskhan: I think what we want is an extra parameter in the UDF for the root. In the unit test you can just specifically pass in the path you get from the fixture.

Since this is actually materializing the images, I'm not sure whether you'll want to make this a local temporary directory. Otherwise any kind of action that causes shuffling/repartitioning will break this dataset.

1. Short term: add uri parameter to udf to address #130 
2. Separate from this PR, we should have a top-level rikai option that specifies a default temp s3 root. In production code the default temporary directory should come from this option
3. In the long-run, @eddyxu what do you think about having something like an "InMemoryImage" or allow `Image` to just have the in memory ndarray?

changhiskhan (changes requested):

changhiskhan:
@smellslikeml we're getting ready to make a v0.1 release tmr. Any chance you could get this wrapped up?

python/rikai/spark/functions/vision.py
smellslikeml: Added a uri parameter for the YouTubeVideo case which defaults to tmp.

python/tests/spark/test_functions.py
```diff
@@ -139,6 +139,7 @@ def test_video_to_images(spark: SparkSession):
     """Test extract video frames from YouTubeVideo/VideoStream types
     into list of Image assets.
     """
+    youtube_uri = os.path.join(os.path.dirname(__file__), "..", "assets")
```
changhiskhan, line 142: this is going to pollute the repo directories. with pytest you can pass in `tmp_dir` into your test function and start using it.  https://docs.pytest.org/en/stable/tmpdir.html

python/rikai/spark/functions/vision.py
```diff
@@ -106,6 +108,7 @@ def numpy_to_image(array: ndarray, uri: str) -> Image:
 def video_to_images(
     video,
     segment: Segment = Segment(0, -1),
+    youtube_uri: str = tempfile.gettempdir(),
```
changhiskhan, line 111: this is actually a python gotcha: https://8thlight.com/blog/shibani-mookerjee/2019/05/07/some-common-gotchas-in-python.html#:~:text=Python's%20default%20arguments%20are%20evaluated,to%20the%20function%20as%20well.

make the default None and then create a tempdir inside the function if it's none

changhiskhan (changes requested):
almost there.

python/tests/spark/test_functions.py
smellslikeml: got it 👍

python/rikai/spark/functions/vision.py
smellslikeml: made the default directory an empty string so we could keep the typing

python/rikai/spark/functions/vision.py
```diff
@@ -122,9 +125,13 @@ def video_to_images(
         The sampling rate in number of frames
     max_samples : Int
         Yield at most this many frames (-1 means no max)
+     youtube_uri: Str
```
changhiskhan, line 128: nit: str is not capitalized

1. The way you have it for VideoStream it's still writing to local disk on each spark executor at the point of invocation. You'll want to make sure that in both YouTubeVideo and VideoStream cases you're writing to the supplied argument (maybe an S3 path in real usage).
2. Instead of a `youtube_uri` argument, you want to call this `base_path` or `output_dir` or something similar. Since that's what it actually is.
3. don't make it optional for now -- the tempdir is cleared as soon as the context exists

python/rikai/spark/functions/vision.py
```diff
@@ -137,13 +144,15 @@ def video_to_images(
     assert isinstance(segment, Segment), "Second input type must be Segment"
 
     base_path = video.uri
+    if not youtube_uri:
+        youtube_uri = tempfile.gettempdir()
```
changhiskhan, line 148: remove this. tempdir auto-clears

python/rikai/spark/functions/vision.py
```diff
@@ -137,13 +144,15 @@ def video_to_images(
...
 
     start_frame = segment.start_fno
     if segment.end_fno > 0:
         max_samples = min((segment.end_fno - start_frame), max_samples)
 
     if isinstance(video, YouTubeVideo):
-        base_path = video.vid
+        base_path = os.path.join(youtube_uri, video.vid)
```
changhiskhan, line 155: prepend the given argument regardless if it's YouTubeVideo or VideoStream

python/rikai/spark/functions/vision.py
eddyxu: I think it is a good time now to think a bit about the user experience of how users interact with the directory to store assets. 

Should we expect a global asset directory (i.e., `rikai.option.set("rikai.asset.dirctory", "s3://bucket/foo/assets"`) or ask users to pass `lit(”s3://bucket/foo/assets")` each time?

changhiskhan:
@smellslikeml i updated this PR to fix some of the issues.

python/rikai/spark/functions/vision.py
eddyxu: @changhiskhan @smellslikeml do you foresee that whether each dataset (the output of one DataFrame) contains one of these asset directory, or there will be a global asset directory for all jobs? 

For example, could it be something like this:

```python

rikai.option.set("rikai.default.asset.directory", "s3://bucket/to/asset")

df.withColumn("image", video_to_image(df.video))  
```

python/rikai/spark/functions/vision.py
```diff
@@ -14,8 +14,10 @@
 
 """Vision related Spark UDFs.
 """
+from typing import Union
 
 # Third Party
+import os
```
eddyxu, line 20: This import should be in the standard library section?

python/rikai/spark/functions/vision.py
```diff
@@ -104,7 +106,8 @@ def numpy_to_image(array: ndarray, uri: str) -> Image:
 
 @udf(returnType=ArrayType(ImageType()))
 def video_to_images(
-    video,
+    video: Union[VideoStream, YouTubeVideo],
+    output_uri: str,
```
eddyxu, line 110: Discussed in another thread,  my main concern is that this parameter drags users from describing algorithms, to must be aware of infra details ("i.e., where the storage is"). This is a distraction to user's workflow. 

I'd encourage us to re-think how Image / Video asset class is implemented.

python/rikai/spark/functions/vision.py
```diff
@@ -158,15 +160,18 @@ def video_to_images(
     return [
         Image.from_array(
             img,
-            "{}_{}.jpg".format(base_path, (start_frame + idx) * sample_rate),
+            os.path.join(
```
eddyxu, line 163: nit, use `pathlib.Path`?

python/tests/spark/test_functions.py
```diff
@@ -135,51 +135,50 @@ def test_numpy_to_image(spark: SparkSession, tmp_path: Path):
...
+    output_dir = tmp_path / "videostream_test"
+    output_dir.mkdir(parents=True)
+    df1 = df1.withColumn(
         "images",
         video_to_images(
-            col("video"), col("segment"), lit(sample_rate), lit(max_samples)
+            col("video"),
+            lit(str(output_dir)),
+            col("segment"),
+            lit(sample_rate),
```
eddyxu, line 158: I am not expect on this. Can pandas UDF parameter actually be optional ?

python/rikai/spark/functions/vision.py
eddyxu: > In the long-run, @eddyxu what do you think about having something like an "InMemoryImage" or allow Image to just have the in memory ndarray?

yea, i think we can make `Image` compatible  to this, basically sth like

```
class Image(UDT):
    def __init__(self):
        self.content = None
        self.uri = None
        # we just keep one of them in the payload.
```

python/rikai/spark/functions/vision.py
changhiskhan: it doesn't work for non-local paths (e.g., it removes one of the slashs from https:// or s3://)

python/rikai/spark/functions/vision.py
changhiskhan: Yeah, I totally agree. Previously we had talked about supporting and Image option that is just an in-memory ndarray and create an explicit externalization method that switches between what we have now for Image and that in-memory representation.
Once we have that, the user can choose to externalize the images, otherwise we store them directly with the image data embedded in the column.

The counter-argument to hiding the storage is that if you automatically stash it somewhere, it'll be very easy for the user to lose the actual data even though they may think they know exactly where they put the dataframe. 

I made #144 to keep track of it. My suggestion here would be to get this change in so this is at least a predictable/consistent feature. Then once we create the alternative Image class we can refactor this one to remove the uri (or make it optional).

python/rikai/spark/functions/vision.py
eddyxu: 👌

python/tests/spark/test_functions.py
changhiskhan: this isn't a pandas udf. Default arguments work fine i believe (see spectrogram test case)?

python/rikai/spark/functions/vision.py
changhiskhan: done

python/rikai/spark/functions/vision.py
eddyxu: Make sense. Lets make things moving :)

eddyxu (approved):

smellslikeml:
The YouTubeVideoType has some [limitations](https://towardsdatascience.com/downloading-the-kinetics-dataset-for-human-action-recognition-in-deep-learning-500c3d50f776) imposed by YouTube but I've found it useful for hacking something out quickly.
Been able to do more free runs on community clusters with easy access to sample data.

Issue #23, opened 2021-01-18
Dataset should support dataframe reference?


changhiskhan:
Currently Dataset only support references to a persisted Rikai dataset's uri. Oftentimes we'll want to load a Dataset, do some transformations, then create the DataLoader for training. It would be inefficient / cumbersome to have to explicitly save the dataset every time.

For now, it would be sufficient to just configure some temporary storage so that in the background it writes a temporary Rikai dataset. Would be interesting in the long-term to explore passing around Arrow buffers in memory as opposed to going through the serde roundtrip.

Issue #134, opened 2021-03-19
Implement a Option class in python


eddyxu:
So that it can back up a few functionalities:

* Implement helping methods like "getInt, getFloat" and etc.
* Manage serialization between JVM and python to pass configurations and options. 

The same class should be helpful for #15 as well.

changhiskhan:
so what do you think about using the pandas options as a temporary `rikai.options` module for now?
Pro: only a few lines of code
Con: someone might accidentally do pd.reset_option('all') and also clear rikai options.

That can be done to satisfy #15 without doing this issue.

What examples did you have in mind that needs this Option class?

eddyxu:
A few places, for example `pre/post` transforms, `model` are using `Dict[str, str]` to inherent options from model specs. With a config/option class, with the supports of `getInt` `getFloat` and etc, it is easier to manage the configurations.

Also this option needs to be passed between python and spark jvm. i.e, model spec parsed in python , registered to JVM's `ModelCatalog`. When we pass the conf/options, now we serialize them into `Map[String, String]`. But when we use it, need to support more types as well.

changhiskhan:
ic. in that case this doesn't seem super urgent. I might do #15 with the pandas hack for now to provide some mechanism for users to override settings easily. That way i can move onto the mlflow stuff sooner.

PR #147, opened 2021-03-26
Create Dataset/DataLoader from DataFrame

changhiskhan:
addresses #23

### DIFF
--- a/python/rikai/spark/utils.py
+++ b/python/rikai/spark/utils.py
@@ -1,0 +1,23 @@
+#  Copyright (c) 2021 Rikai Authors
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from pyspark.sql import DataFrame
+
+DEFAULT_ROW_GROUP_SIZE_BYTES = 32 * 1024 * 1024
+
+def df_to_parquet(df: DataFrame, uri: str,
+        parquet_row_group_size_bytes: int=DEFAULT_ROW_GROUP_SIZE_BYTES):
+    (df.write.option("parquet.block.size", parquet_row_group_size_bytes)
+    .parquet(uri))
+
--- a/python/rikai/spark/utils.py
+++ b/python/rikai/spark/utils.py
@@ -16,8 +16,14 @@ from pyspark.sql import DataFrame
 
 DEFAULT_ROW_GROUP_SIZE_BYTES = 32 * 1024 * 1024
 
-def df_to_parquet(df: DataFrame, uri: str,
-        parquet_row_group_size_bytes: int=DEFAULT_ROW_GROUP_SIZE_BYTES):
-    (df.write.option("parquet.block.size", parquet_row_group_size_bytes)
-    .parquet(uri))
 
+def df_to_rikai(
+    df: DataFrame,
+    uri: str,
+    parquet_row_group_size_bytes: int = DEFAULT_ROW_GROUP_SIZE_BYTES,
+):
+    (
+        df.write.format("rikai")
+        .option("parquet.block.size", parquet_row_group_size_bytes)
+        .save(uri)
+    )
--- a/python/rikai/torch/data.py
+++ b/python/rikai/torch/data.py
@@ -13,12 +13,13 @@
 #  limitations under the License.
 
 """Pytorch Dataset and DataLoader"""
-
+import os
 import threading
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
 from queue import Empty, Queue
 from typing import Callable, Dict, Generator, List, Optional, Union
+import uuid
 
 # Third Party
 import torch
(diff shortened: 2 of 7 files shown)


python/rikai/torch/data.py
```diff
@@ -70,17 +75,19 @@ class Dataset(IterableDataset):
...
         super().__init__()
-        self.uri = str(uri)
+        self.uri_or_df = uri_or_df
         self.columns = columns
         self._transform = transform
+        self.parquet_row_group_size_bytes = parquet_row_group_size_bytes
 
     def __repr__(self) -> str:
-        return f"Dataset(torch, {self.uri}, columns={self.columns})"
+        return f"Dataset(torch, {self.uri_or_df}, columns={self.columns})"
```
eddyxu, line 90: nit, is df printable with meaningful info here?

python/rikai/spark/utils.py
```diff
@@ -0,0 +1,23 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from pyspark.sql import DataFrame
+
+DEFAULT_ROW_GROUP_SIZE_BYTES = 32 * 1024 * 1024
+
+def df_to_parquet(df: DataFrame, uri: str,
+        parquet_row_group_size_bytes: int=DEFAULT_ROW_GROUP_SIZE_BYTES):
```
eddyxu, line 20: We really need #134  :)

python/rikai/torch/data.py
```diff
@@ -254,3 +268,55 @@ def __iter__(self) -> Generator[Dict, None, None]:
...
+    else:
+        from pyspark.sql import DataFrame
+        from rikai.spark.utils import df_to_parquet
+        if isinstance(dataset_ref, DataFrame):
+            kwargs = {}
+            if parquet_row_group_size_bytes is not None:
+                kwargs['parquet_row_group_size_bytes'] = \
+                    parquet_row_group_size_bytes
+            _get_cache_uri(dataset_ref)
+            df_to_parquet(dataset_ref, cache_uri, **kwargs)
```
eddyxu, line 303: raise if it is not a dataframe, or pyspark is not imported / installed?

python/rikai/torch/data.py
```diff
@@ -167,7 +176,12 @@ def __init__(
         seed: Optional[int] = None,
         world_size: int = 1,
         rank: int = 0,
+        parquet_row_group_size_bytes: int = None
```
eddyxu, line 179: maybe put this into either SparkSession.ctx or `rikai.option`? 

This is a writer site parameter, which should not on `Dataset`

eddyxu (commented):
Generally i think we are aligned.

python/rikai/torch/data.py
changhiskhan: this pyspark DataFrame's repr:

```
    def __repr__(self):
        if not self._support_repr_html and self.sql_ctx._conf.isReplEagerEvalEnabled():
            vertical = False
            return self._jdf.showString(
                self.sql_ctx._conf.replEagerEvalMaxNumRows(),
                self.sql_ctx._conf.replEagerEvalTruncate(), vertical)
        else:
            return "DataFrame[%s]" % (", ".join("%s: %s" % c for c in self.dtypes))
```

python/rikai/torch/data.py
changhiskhan: added more checks

python/rikai/spark/utils.py
changhiskhan: for now i'm just going to use spark conf

python/rikai/spark/utils.py
```diff
@@ -0,0 +1,29 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from pyspark.sql import DataFrame
+
+DEFAULT_ROW_GROUP_SIZE_BYTES = 32 * 1024 * 1024
```
eddyxu, line 17: nit: we should consolidate these const / conf into one center place.  no blocker

python/rikai/torch/data.py
```diff
@@ -40,10 +41,12 @@ class Dataset(IterableDataset):
 
     Parameters
     ----------
-    uri : str
+    uri_or_df : str, Path, pyspark.sql.DataFrame
```
eddyxu, line 44: do we want to be consistent between `uri_or_df` and `dataset_ref`?
I like `dataset_ref` more , it can be used for "query" as well.

python/rikai/torch/data.py
```diff
@@ -13,12 +13,13 @@
...
 
 """Pytorch Dataset and DataLoader"""
-
+import os
 import threading
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
 from queue import Empty, Queue
 from typing import Callable, Dict, Generator, List, Optional, Union
+import uuid
```
eddyxu, line 22: `isort` imports?

eddyxu (approved):
Small nits

python/rikai/torch/data.py
changhiskhan: I'm going to rename it `data_ref` throughout (i like dataset_ref better as well but it's weird to pass `dataset_ref` to the `Dataset` constructor. I don't want the users to get confused thinking `dataset_ref` is a reference to another instance of `rikai.torch.data.Dataset`.

python/rikai/spark/utils.py
changhiskhan: adding a `rikai.spark.conf` module

python/rikai/torch/data.py
changhiskhan: massive PR coming :)

Issue #20, opened 2021-01-16
Support Google Cloud Storage


eddyxu:
To run to GCP

PR #149, opened 2021-03-27
Support GCS in rikai IO

eddyxu:
Support Google cloud storage in Rikai IO for #20

### DIFF
--- a/python/rikai/io.py
+++ b/python/rikai/io.py
@@ -19,6 +19,7 @@ from os.path import basename, join
 from pathlib import Path
 from typing import IO, BinaryIO, Union
 from urllib.parse import ParseResult, urlparse
+import functools
 
 # Third Party
 import requests
@@ -28,8 +29,6 @@ from pyarrow import fs
 from rikai.logging import logger
 
 __all__ = ["copy", "open_uri"]
-
-_GCSFS = None
 
 
 def _normalize_uri(uri: str) -> str:
@@ -49,17 +48,17 @@ def _normalize_uri(uri: str) -> str:
     ).geturl()
 
 
-def _gcsfs():
+@functools.lru_cache(maxsize=1)
+def _gcsfs(project="", token=None, block_size=None):
     try:
         import gcsfs
     except ImportError as e:
         raise ImportError(
             "Please make sure gcsfs is installed via `pip install rikai[gcp]`"
         ) from e
-    global _GCSFS
-    if _GCSFS is None:
-        _GCSFS = gcsfs.GCSFileSystem()
-    return _GCSFS
+    return gcsfs.GCSFileSystem(
+        project=project, token=token, block_size=block_size
+    )
 
 
 def _open_input_stream(uri: str) -> BinaryIO:
--- a/python/rikai/io.py
+++ b/python/rikai/io.py
@@ -13,13 +13,13 @@
 #  limitations under the License.
 
 # Standard
+import functools
 import shutil
 from io import BytesIO
 from os.path import basename, join
 from pathlib import Path
 from typing import IO, BinaryIO, Union
 from urllib.parse import ParseResult, urlparse
-import functools
 
 # Third Party
 import requests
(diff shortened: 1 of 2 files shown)


eddyxu:
Have tested manually like

```python
from rikai.io import copy
from rikai.types.vision import Image

copy("local.png", "gs://my_bucket/")
copy("gs://my_bucket/local.png", "gs://my_bucket/local_copy.png")

print(Image("gs://my_bucket/local.png").to_numpy())
```

We dont have the infra to test this in rikai yet.

python/setup.py
```diff
@@ -7,8 +7,7 @@
     exec(fh.read(), about)
 
 with open(
-    pathlib.Path(__file__).absolute().parent.parent / "README.md",
-    "r",
+    pathlib.Path(__file__).absolute().parent.parent / "README.md", "r"
```
eddyxu, line 10: This was re-formatted by `black`

python/rikai/io.py
```diff
@@ -76,17 +109,19 @@ def copy(source: str, dest: str) -> str:
...
 
     # TODO: find better i/o utilis to copy between filesystems
-    filesystem, dest_path = fs.FileSystem.from_uri(dest)
-    with filesystem.open_output_stream(dest_path) as out_stream:
-        src_fs, src_path = fs.FileSystem.from_uri(source)
-        with src_fs.open_input_stream(src_path) as in_stream:
+    with _open_output_stream(dest) as out_stream:
+        with _open_input_stream(source) as in_stream:
             while True:
                 buf = in_stream.read(_BUFSIZE)
```
chunyang, line 126: Can probably use `shutil` here instead of implementing it ourselves.

python/rikai/io.py
```diff
@@ -49,6 +49,39 @@ def _normalize_uri(uri: str) -> str:
     ).geturl()
 
 
+def _gcsfs():
```
chunyang, line 52: Use `@functools.cache` instead of maintaining the `_GCSFS` global variable?

python/rikai/io.py
```diff
@@ -124,6 +159,8 @@ def open_uri(uri: Union[str, Path], mode="rb") -> BinaryIO:
     elif parsed_uri.scheme in ("http", "https"):
         resp = requests.get(uri)
         return BytesIO(resp.content)
+    elif parsed_uri.scheme == "gs":
+        return _gcsfs().open(uri, mode=mode)
```
chunyang, line 163: I think the type signature is wrong? Depending on `mode`, the return value is not always a `BinaryIO`.

chunyang (approved):
Code LGTM, though I think one of the type signatures is wrong and there are a few places where code code be simplified with use of the standard library.

python/rikai/io.py
eddyxu: Can shutil work with cloud storage?

python/rikai/io.py
eddyxu: TIL

python/rikai/io.py
eddyxu: Hmm, good point.

python/rikai/io.py
chunyang: I think [`shutil.copyfileobj`](https://docs.python.org/3/library/shutil.html#shutil.copyfileobj) can work with any file-like object.

python/rikai/io.py
eddyxu: `@functools.cache` is new from python 3.9...
We need to provide compatibility from 3.7 tho.

https://docs.python.org/3/library/functools.html

Other alternatives?

python/rikai/io.py
eddyxu: `shutil.copyfileobj` ?

python/rikai/io.py
eddyxu: guess we can use `lru`  🤫

PR #155, opened 2021-03-27
Add warnings to deprecate rikai.torch.DataLoader

eddyxu:
After introducing `BufferedShuffleDataset` in Pytorch 1.8,  the official DataLoader can shuffle an IterableDataset (a.k.a, Rikai's parquet Dataset). Therefore, we dont need implement DataLoader by ourselves.

This PR starts to show warnings to alert users to use torch's DataLoader if they are using pytorch 1.8+

### DIFF
--- a/python/rikai/torch/data.py
+++ b/python/rikai/torch/data.py
@@ -15,12 +15,14 @@
 """Pytorch Dataset and DataLoader"""
 
 import threading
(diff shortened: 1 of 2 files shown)


changhiskhan (approved):
lol, deprecation warning pre-0.1 is kind of funny

Issue #144, opened 2021-03-25
Create in-memory/embedded version of Image class


changhiskhan:
See discussion here https://github.com/eto-ai/rikai/pull/133/files#r601792580

changhiskhan:
duplicate of #143

PR #156, opened 2021-03-28
FFmpeg scene change detection

smellslikeml:
FFmpeg filters threshold [sum of absolute differences](https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/f_select.c) in video frames to perform scene change detection.

This can be used to [featurize](https://blog.gdeltproject.org/using-ffmpegs-scene-detection-to-generate-a-visual-shot-summary-of-television-news/) and sample video, especially efficient as sparse representation of slowly varying video.

With complex, multi-camera videos, pipelines benefit from scene detection to decompose video into Segments as contiguous sequences of video frames, where simplifying assumptions for object tracking hold.

image

Relates to projects like [pyscenedetect](https://pyscenedetect.readthedocs.io/en/latest/)


python/rikai/spark/functions/vision.py
```diff
@@ -234,3 +237,79 @@ def spectrogram_image(
...
+        except ValueError:
+            return False
+
+    video_uri = (
+        video.get_stream().uri
+        if isinstance(video, YouTubeVideo)
+        else video.uri
+    )
+
+    p = subprocess.Popen(
```
changhiskhan, line 289: if we're going to start adding multiple ffmpeg related UDFs, we should think about adding a fluent API. Something that might look something like:

```
result = rikai.video.ffmpeg.run_pipline(
    ffmpeg.input(video_uri)
    .filter("select", "gte(scene,{})".format(threshold))
    .filter("showinfo")
    .output("-", format="null")
)
```

some of the type conversion tools can also be common utilities since all ffmpeg pipelines need to interact the cmdline subprocess

python/rikai/spark/functions/vision.py
```diff
@@ -234,3 +237,79 @@ def spectrogram_image(
...
+            .filter("select", "gte(scene,{})".format(threshold))
+            .filter("showinfo")
+            .output("-", format="null")
+            .compile()
+        ),
+        stderr=subprocess.PIPE,
+    )
+    result = p.communicate()[1].decode("utf-8")
+    cut_times = [0] + [
+        float(ln.split()[0])
```
changhiskhan, line 301: what does the output look like when it's not a float?

python/rikai/spark/functions/vision.py
```diff
@@ -234,3 +237,79 @@ def spectrogram_image(
...
+        (
+            ffmpeg.input(video_uri)
+            .filter("select", "gte(scene,{})".format(threshold))
+            .filter("showinfo")
+            .output("-", format="null")
+            .compile()
+        ),
+        stderr=subprocess.PIPE,
+    )
+    result = p.communicate()[1].decode("utf-8")
```
changhiskhan, line 299: what happens if the command fails?

python/rikai/spark/functions/vision.py
```diff
@@ -234,3 +237,79 @@ def spectrogram_image(
...
+
+
+@udf(returnType=ArrayType(SegmentType()))
+def video_segmentor(
+    video: Union[VideoStream, YouTubeVideo],
+    threshold: float = 0.3,
+    fps: int = 30,
+) -> list:
+    """Returns list of video Segments from VideoStream or YouTubeVideo by
+    thresholding sum-of-absolute differences.
```
changhiskhan, line 249: Are there popular alternatives to sum-of-absolute differences? If so, how might we make this configurable so we can use other methods?

changhiskhan (commented):
@eddyxu what do you think about the Segment definition comment? Do you see it more common to define the segment by frame numbers or times?

python/rikai/spark/functions/vision.py
```diff
@@ -234,3 +237,79 @@ def spectrogram_image(
...
+        if isfloat(ln.split()[0])
+    ]
+    if len(cut_times) == 1:
+        return [Segment(0, -1)]
+    intervals = [
+        (cut_times[idx - 1], cut_times[idx])
+        for idx, time in enumerate(cut_times)
+    ][1:]
+    return [
+        Segment(start_fno=int(fps * tup[0]), end_fno=int(fps * tup[1]))
```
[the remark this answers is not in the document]
changhiskhan: i guess the framerate probe might help, but then we should think about how to keep the framerate info preserved. e.g., maybe we have 2 ways to express segments: `Segment(start_fno, end_fno, framerate)` and `Segment(start_time, end_time)` if the framerate isn't handy (or if we need it to be portable across multiple versions of the video)

python/rikai/spark/functions/vision.py
smellslikeml: The output is a float cast as a string e.g. `'23.47'`

python/rikai/spark/functions/vision.py
smellslikeml: will find some test cases but I expect the output in result will not have `pts_time:` used to parse timestamps

python/rikai/spark/functions/vision.py
smellslikeml: [ffmpeg's](http://www.ffmpeg.org/ffmpeg-filters.html#select_002c-aselect) scene detection uses simple threshold based detection without much flexibility.

[pyscenedetect](https://pyscenedetect.readthedocs.io/en/latest/reference/detection-methods/) offers a similar threshold based scene detector in addition to a more sensitive content-aware detector. We could use their base `SceneDetector` to make custom ones.

changhiskhan: do you think that ffmpeg overall is just too low-level of a tool for us to use directly in rikai? feels like in at least two things, OpenCV and pyscenedetect there are higher level APIs that are available?

changhiskhan:
#177 will load the whole video into memory. And videos that need scene detection is probably not short. are you not concerned about blowing up the cluster memory?

smellslikeml:
> #177 will load the whole video into memory. And videos that need scene detection is probably not short. are you not concerned about blowing up the cluster memory?

Scene detection will take place without loading the video and help to narrow the scope down to Segments.

We could limit to "interesting" segments using a method to sample a few frames for inference.

My goal would be to use FFmpeg to reduce the video as much as needed in space/time, with limited intermediary representations.

python/rikai/spark/functions/vision.py
smellslikeml: low-level like a fox! FFmpeg can help to decode all kinds of streams from multimedia containers.

I also found that ffmpeg can be built to run [opencv as filters](http://underpop.online.fr/f/ffmpeg/help/ocv.htm.gz).

smellslikeml: Finding relevant tensorflow-io pulls on this topic:

[#11 ](https://github.com/tensorflow/io/issues/11)
[#144](https://github.com/tensorflow/io/issues/144)
[#30](https://github.com/tensorflow/io/pull/30)

smellslikeml: maybe something like this [example](https://github.com/kkroening/ffmpeg-python/blob/master/examples/tensorflow_stream.py) that produces images from bytes without converting to numpy.

smellslikeml: [skvideo scene detection](http://www.scikit-video.org/stable/examples/scene.html#sceneexample)

PR #162, opened 2021-03-29
Create python-publish.yml

changhiskhan:
Publish to PyPI once we make a release in GH

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,33 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      working-directory: python
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build and publish
+      working-directory: python
+      env:
+        TWINE_USERNAME: __token__
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,33 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
```
eddyxu, line 8: Is this the same as "filter based on tag"?

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
```
eddyxu, line 20: Nit: might want to add document about the procedure to create tags. 
For example, do we need a green build of the commit before we push a tag? Or we will include the test in publish as well?

cc @chunyang has many experience on this as well.

.github/workflows/python-publish.yml
changhiskhan: no this is us manually creating a "release" in github.
https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository

.github/workflows/python-publish.yml
changhiskhan: probably wouldn't hurt to run the tests again just in case.

eddyxu: From what I see yesterday is that, a release in python needs to match a *Released* version of jar on maven central. For example, `0.0.3` release needs to wait maven central to propagate `rikai_0.12:0.0.3` jar being published. Before that even the good tests will fail.  (or they are passing but it means that we did not update version number)

changhiskhan: oh but that complicates the setup and you'll have to install java/scala/sbt etc also

changhiskhan: `From what I see yesterday is that, a release in python needs to match a Released version of jar on maven central.` for our users yes that's true. The issue is waiting on maven central propagation isn't deterministic. We could test with doing `sbt publishLocal` to not require this wait time tho

eddyxu: >oh but that complicates the setup and you'll have to install java/scala/sbt etc also
right, that was why i ask whether we wanted to run tests in this github action. Or we can just release as it is.

eddyxu (approved):
LGTM.

.github/workflows/python-publish.yml
changhiskhan: we *can* run tests (via publishLocal) if we needed to, just makes the setup more complicated. And as long as the commit is green then technically there's no need

PR #166, opened 2021-03-29
Create scala-publish

changhiskhan:
workflow based on sbt-ci-release template

note that the template (https://github.com/olafurpg/sbt-ci-release/blob/main/.github/workflows/release.yml)
also publishes SNAPSHOT artifacts whereas this PR does not

### DIFF
--- a/.github/workflows/scala-publish
+++ b/.github/workflows/scala-publish
@@ -1,0 +1,20 @@
+name: Publish Jar to Sonatype
+on:
+  release:
+    types: [created]
+    
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 0
+      - uses: olafurpg/setup-scala@v10
+      - uses: olafurpg/setup-gpg@v3
+      - run: sbt ci-release
+        env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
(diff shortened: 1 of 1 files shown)


.github/workflows/scala-publish
```diff
@@ -0,0 +1,20 @@
+name: Publish Jar to Sonatype
+on:
+  release:
+    types: [created]
```
eddyxu, line 4: guess my question is that how these test can fail if we did not bump version number from `0.0.4-SNAPSHOT` to `0.0.4` ?

nit tho.

eddyxu (approved):

.github/workflows/scala-publish
changhiskhan: So the idea is that we'll use `bump2version` to increment the python version from say 0.0.4dev0 to 0.0.4 and create a v0.0.4 tag. sbt dynver will then use the tag to determine that this is 0.0.4 jar (non-snapshot since it's tagged). 

For conftest and notebooks, I will add something else to automatically update those references.

The main way this could go wrong is if we create a release on a commit without a tag (e.g., we forgot to add a tag, we pointed to the wrong commit, we forgot to bump version and somehow added the tag manually). That would cause sbt to publish a snapshot build to the staging repo and pypi to get a dev0 build. In that scenario I believe we can just fix master and then create another release. It would re-trigger the two release workflows.

Another possibility is if we didn't wait for the tagged commit to be green and it failed later. In that case we may have to just bump the version again and mark a particular release as broken or something

PR #170, opened 2021-03-30
[TEST] Add more test timeouts

changhiskhan:
Mainly to check if downloading fasterrcnn_resnet50 is the culprit

### DIFF
--- a/python/tests/spark/sql/codegen/test_fs.py
+++ b/python/tests/spark/sql/codegen/test_fs.py
@@ -14,6 +14,7 @@
 
 import tempfile
 from pathlib import Path
+import uuid
 
 import pytest
 import torch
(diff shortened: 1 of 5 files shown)


python/tests/spark/sql/codegen/test_fs.py
```diff
@@ -33,37 +34,36 @@
...
-    with tempfile.TemporaryDirectory() as tmp_dir:
-        tmp_path = Path(tmp_dir)
-        # Prepare model
-        resnet = torchvision.models.detection.fasterrcnn_resnet50_fpn(
-            pretrained=True,
-            progress=False,
-        )
-        model_uri = tmp_path / "resnet.pth"
-        torch.save(resnet, model_uri)
+    tmp_path = tmp_path_factory.mktemp(str(uuid.uuid4()))
```
eddyxu, line 40: just curious, is this tmp_path being tracked and auto-deleted by `tmp_path_factory`?

eddyxu (approved):

python/tests/spark/sql/codegen/test_fs.py
changhiskhan: yeah, it's session scoped

PR #171, opened 2021-03-30
add utility to automatically get the matching jar version from python version

changhiskhan:
i'll update the release docs after we make another patch release or two and things get settled


notebooks/Coco.ipynb
```diff
@@ -18,14 +18,16 @@
     "from pyspark.sql import SparkSession\n",
     "import pyspark.sql.functions as F\n",
     "import numpy as np\n",
+    "from rikai.spark.utils import get_default_jar_version\n",
     "\n",
+    "version = get_default_jar_version(False)\n",
```
eddyxu, line 23: should we use `get_dfeault_jar_version(use_snapshot=False)` to make it more readable?

eddyxu: if we dont use snapshot version in notebook, will it already fail?
For example, it is `0.0.4-snapshot` at the moment, this notebook will try to include `0.0.4` which has not been released?

changhiskhan: it will fail

eddyxu: doesnt that mean if anyone open a notebook from master branch, the notebook will not work?

changhiskhan: yeah, let me update this. I broke this out from the sbt-ci-release change but i forgot i manually bumped the dynver patch version number :(

changhiskhan: i think just setting `use_snapshot=True` would fix the problem

eddyxu (approved):

PR #173, opened 2021-03-31
Remove boto from python dependency.

eddyxu:
For now, we dont need `boto` as dependency. All s3 accesses are done via `pyarrow`.


changhiskhan (approved):
Assuming aws access still works locally (since we don't have integration tests yet)?

eddyxu:
Yes, it works. I did not ever install `boto` actually.

Issue #143, opened 2021-03-25
Make Asset class to support both in-place data or external data


eddyxu:
We could make Asset class more flexible to support both in-place data or external data.

For example, for certain assets (i.e., small images), it might make sense that only keep the data within `parquet` columns. 

```
class Image(Asset):
    def __init__(self):
        # we just keep one of them in the payload.
        self.content = None
        self.uri = None
```

PR #175, opened 2021-03-31
To from bytestring

smellslikeml:
spark's image format uses bytestrings to represent images.

image

Adding methods to read images into and from bytestrings will help in loading from parquet formatted this way.

image


python/rikai/types/vision.py
```diff
@@ -104,6 +105,20 @@ def from_pil(img: PILImage, uri: Union[str, Path]) -> Image:
...
+        """Create an image in memory from a byte string.
+
+        Parameters
+        ----------
+        img : bytes
+            An PIL Image instance
+        uri : str or Path
+            The URI to store the image externally.
+        """
+        img = PILImage.open(BytesIO(img))
```
changhiskhan, line 119: if this is returning an Image then you gotta close your resources.

```
with PILImage.open(BytesIO(img_bytes)) as pil_img:
    return self.from_pil(pil_img, uri)
```

python/rikai/types/vision.py
```diff
@@ -163,3 +178,16 @@ def to_numpy(self) -> np.ndarray:
...
+
+    def to_bytes(self) -> bytes:
+        """Return a byte array.
+
+        Note
+        ----
+        The caller should close the image.
+        https://pillow.readthedocs.io/en/stable/reference/open_files.html#image-lifecycle
+        """
+        img = PILImage.open(self.open())
```
changhiskhan, line 190: same here with closing resources.

```
with self.to_pil() as img:
    img_bytes = BytesIO()
    img.save(img_bytes, format='PNG')
    return img_bytes.getvalue()
```

why PNG format as default? This is probably something that we should cleanup later and parameterize the image format.

changhiskhan (commented):
1. Is the idea that we want to convert from Spark's Imge Data Source? If so, maybe there's a more efficient way to do this without the round-trip conversion?
2. Can you show me the relevant code context around this? If it's an internal databricks notebook just PM me on Slack?

python/rikai/types/vision.py
smellslikeml: good catch, thanks!

python/rikai/types/vision.py
smellslikeml: just went this way when 'JPEG/JPG' failed but I see now that I needed to specify the format in [lowercase](https://pillow.readthedocs.io/en/5.1.x/handbook/image-file-formats.html#jpeg)

python/rikai/types/vision.py
```diff
@@ -104,6 +105,20 @@ def from_pil(img: PILImage, uri: Union[str, Path]) -> Image:
                 copy(fobj.name, uri)
         return Image(uri)
 
+    @staticmethod
+    def from_bytes(img: bytes, uri: Union[str, Path]) -> Image:
```
eddyxu, line 109: Wonder what would this interface looks like if we have #143

python/rikai/types/vision.py
```diff
@@ -104,6 +105,20 @@ def from_pil(img: PILImage, uri: Union[str, Path]) -> Image:
...
+        """Create an image in memory from a byte string.
+
+        Parameters
+        ----------
+        img : bytes
+            An PIL Image instance
```
eddyxu, line 115: Would like to confirm that the input are bytes, not a PIL instance?

eddyxu:
In general I have a little bit concern that this loads image into bytes from Spark Image? and then save another copy to external storage just to make it workable with `Image` class.

python/rikai/types/vision.py
smellslikeml: thanks, updated.

smellslikeml:
> In general I have a little bit concern that this loads image into bytes from Spark Image? and then save another copy to external storage just to make it workable with `Image` class. We might want to implement #143 to unblock this feature.

#143 sounds good to me. I've had issues reading many small image files and found serialization with bytestrings into parquet efficient for loading large collections.

python/rikai/types/vision.py
changhiskhan: the bytes will differ per image format right? Maybe add in format/kwargs like #178 
see PIL docs for details: https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save

changhiskhan:
also the build seems to fail here. please make sure you run `make fix` and also `pytest` locally

smellslikeml:
Closing this PR after recent changes to `ImageType`, we can update `rikai.spark.functions.image` to read from bytes.

PR #177, opened 2021-04-02
FFmpeg load video

smellslikeml:
FFmpeg probe, load video and other utilities.

[Related Discussion](https://github.com/eto-ai/rikai/pull/156#discussion_r605313446)
[Related PR](https://github.com/eto-ai/rikai/pull/157)


python/rikai/video/ffmpeg.py
```diff
@@ -0,0 +1,116 @@
...
+    raise ValueError(
+        "Couldn't import ffmpeg. Please make sure to "
+        "`pip install ffmpeg-python` explicitly or install "
+        "the correct extras like `pip install rikai[all]`"
+    )
+from rikai.types.vision import Image
+from rikai.types.video import VideoSampler
+
+
+class VideoFrameSampler(VideoSampler):
```
changhiskhan, line 32: Does this replace SingleFrameSampler using OpenCV? Seems like these are duplicates?

python/rikai/video/ffmpeg.py
```diff
@@ -0,0 +1,116 @@
...
+                if self.max_samples
+                else None,
+                r=1 / self.sample_rate,
+            )
+            .run(capture_stdout=True)
+        )
+        h = self.height if not self.scale_height else self.scale_height
+        w = self.height if not self.scale_width else self.scale_width
+
+        video_array = np.frombuffer(video_data, np.uint8).reshape(
```
changhiskhan, line 109: one thing here it seems like with OpenCV it's lazily pulling data when iterating but here you're pulling the whole video into memory prior to iterating.

python/rikai/video/ffmpeg.py
```diff
@@ -0,0 +1,116 @@
...
+        self.scale_width = (scale_width,)
+        self.scale_height = (scale_height,)
+        self.probe = None
+
+    def _probe(self):
+        try:
+            self.probe = ffmpeg.probe(self.stream.uri)
+        except ffmpeg.Error as e:
+            print(e.stderr, file=sys.stderr)
+            sys.exit(1)
```
changhiskhan, line 69: ?

python/rikai/video/ffmpeg.py
```diff
@@ -0,0 +1,116 @@
...
+        self.max_samples = max_samples
+        self.scale_width = (scale_width,)
+        self.scale_height = (scale_height,)
+        self.probe = None
+
+    def _probe(self):
+        try:
+            self.probe = ffmpeg.probe(self.stream.uri)
+        except ffmpeg.Error as e:
+            print(e.stderr, file=sys.stderr)
```
changhiskhan, line 68: you'll want to use a logger instead of bare printing (see logger usage in other parts of the code base for examples). Log at error level. Also, I don't know if you'll want to log the entire ffmpeg output -- is there a way to just get a message + stacktrace?

python/rikai/video/ffmpeg.py
changhiskhan: Also with OpenCV you can hook it up to a live stream uri and be able to sample frames from it.

changhiskhan (commented):
Maybe better if we zoom early next week for like 30 min and figure out the right strategy here. I think what you're trying to build with this series of PRs is valuable, but we need to build it in a way such that it's generalizable and scalable.

python/rikai/video/ffmpeg.py
smellslikeml: Capturing frames from live stream is a great use case for OpenCV over the FFmpeg.

python/rikai/video/ffmpeg.py
smellslikeml: Modeled this from `SingleFrameSampler` in case we wanted to load all frames or iterate.

python/rikai/video/ffmpeg.py
smellslikeml: From [here](https://github.com/kkroening/ffmpeg-python/blob/4cb7d26f55208b01b123f07b13108bf4143794e2/examples/video_info.py)

python/rikai/video/ffmpeg.py
smellslikeml: Found something related while looking into SSIM:
image

python/rikai/video/ffmpeg.py
smellslikeml: An interesting [example](https://stonesoup.readthedocs.io/en/v0.1b4/stonesoup.reader.html#stonesoup.reader.video.FFmpegVideoStreamReader) which captures from RTSP stream using FFmpeg

changhiskhan:
@da-tubi any features from this PR worth bringing into spark-video? If not let's close this PR for now

darcy-shen:
cropWidth/cropHeight could be implemented in spark-video

Let us close it.

PR #178, opened 2021-04-02
Pass through image options when saving images

changhiskhan:
When we save images via PIL, there are lots of ways to parametrize.
https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save

### DIFF
--- a/python/rikai/spark/functions/vision.py
+++ b/python/rikai/spark/functions/vision.py
@@ -73,7 +73,9 @@ def image_copy(img: Image, uri: str) -> Image:
 
 
 @udf(returnType=ImageType())
-def numpy_to_image(array: ndarray, uri: str) -> Image:
+def numpy_to_image(
+    array: ndarray, uri: str, format: str = None, **kwargs
+) -> Image:
     """Convert a numpy array to image, and upload to external storage.
 
     Parameters
@@ -82,6 +84,11 @@ def numpy_to_image(array: ndarray, uri: str) -> Image:
         Image data.
     uri : str
         The base directory to copy the image to.
+    format : str, optional
+        The image format to save as. See
+        `supported formats <https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save>`_ for details.
+    kwargs : dict, optional
+        Optional arguments to pass to `PIL.Image.save <https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save>`_.
 
     Return
     ------
@@ -102,7 +109,7 @@ def numpy_to_image(array: ndarray, uri: str) -> Image:
     --------
     :py:meth:`rikai.types.vision.Image.from_array`
     """
-    return Image.from_array(array, uri)
+    return Image.from_array(array, uri, format=format, **kwargs)
 
 
 @udf(returnType=ArrayType(ImageType()))
@@ -113,6 +120,8 @@ def video_to_images(
     sample_rate: int = 1,
     max_samples: int = 15000,
     quality: str = "worst",
+    img_format: str = "png",
+    **img_kwargs,
 ) -> list:
     """Extract video frames into a list of images.
 
@@ -121,7 +130,7 @@ def video_to_images(
     video : Video
         An video object, either YouTubeVideo or VideoStream.
     output_uri: str
-        Frames will be written as <output_uri>/<fno>.jpg
+        Frames will be written as <output_uri>/<fno>.<img_format>
     segment: Segment, default Segment(0, -1)
         A Segment object, localizing video in time to (start_fno, end_fno)
     sample_rate : int, default 1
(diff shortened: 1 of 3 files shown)


python/rikai/types/vision.py
```diff
@@ -93,13 +102,18 @@ def from_pil(img: PILImage, uri: Union[str, Path]) -> Image:
             An PIL Image instance
         uri : str or Path
             The URI to store the image externally.
+        format : str, optional
+            The image format to save as. See
+            `supported formats <https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save>`_ for details.
```
changhiskhan, line 107: @eddyxu do you see a difference between this line and line 78 above? For some reason the linter likes that line but doesn't like this one...

python/rikai/spark/functions/vision.py
```diff
@@ -113,6 +120,8 @@ def video_to_images(
     sample_rate: int = 1,
     max_samples: int = 15000,
     quality: str = "worst",
+    img_format: str = "png",
```
eddyxu, line 123: should we either name this as `format` or `image_format`?

while `format` is reserved word, it is not worse than what PIL takes ?

python/rikai/types/vision.py
```diff
@@ -59,6 +59,8 @@ def from_array(
         array: np.ndarray,
         uri: Union[str, Path],
         mode: str = None,
+        format: str = None,
```
eddyxu, line 62: yea, here it uses `format`.  Lets be consist with `format` and `**kwargs`?

python/rikai/spark/functions/vision.py
eddyxu: Additionally, can we test them in the `pyspark.DataFrame` ,and in `spark sql` format?

I'd like to see the actual UX spark sql.

python/tests/types/test_vision.py
```diff
@@ -44,6 +45,29 @@ def test_show_embedded_jpeg(tmp_path):
...
+
+    expected_uri = str(tmp_path / "expected.jpg")
+    PILImage.fromarray(rescaled).save(
+        expected_uri, format="jpeg", optimize=True
+    )
+
+    assert filecmp.cmp(result_uri, expected_uri)
+
+    result_uri = str(tmp_path / "result.png")
+    Image.from_array(rescaled, result_uri, format="png", compress_level=1)
```
eddyxu, line 62: `Image.from_array` can take `Path` now.

python/rikai/types/vision.py
```diff
@@ -93,13 +102,18 @@ def from_pil(img: PILImage, uri: Union[str, Path]) -> Image:
...
+        format : str, optional
+            The image format to save as. See
+            `supported formats <https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save>`_ for details.
+        kwargs : dict, optional
+            Optional arguments to pass to `PIL.Image.save <https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save>`_.
+        """  # noqa: E501
         parsed = urlparse(normalize_uri(uri))
         if parsed.scheme == "file":
-            img.save(uri)
+            img.save(uri, format=format, **kwargs)
```
eddyxu, line 113: Just curious, what is the implication that we just keep `format` in `**kwargs`?

python/rikai/spark/functions/vision.py
changhiskhan: naming: i wanted to distinguish the video options from the image options. Here there are many parameters so it's easy for users to get confused.

in DataFrame format with UDF you can safely omit the arguments for optional parameters. If you need to specify these then it's just adding `img_format=lit("png")` no?

SparkSQL - i can check but i don't think it's a problem either?

python/rikai/types/vision.py
changhiskhan: this is API for the image itself and keeping consistency with PIL.Image makes sense, but I think we don't necessarily need that for the video_to_image function

python/tests/types/test_vision.py
changhiskhan: forgot. i can change the test case

python/rikai/types/vision.py
changhiskhan: PIL.Image separates it out so i just did it for consistency. 🤷

python/rikai/spark/functions/vision.py
changhiskhan: wrt `image_format` I guess I prefer brevity where the abbreviation is obvious. I can change it to that given it might not be obvious to all.

changhiskhan (commented):
in test_functions.py you can see examples of using DataFrame API to call UDF with optional arguments.

python/rikai/types/vision.py
```diff
@@ -81,10 +88,12 @@ def from_array(
 
         assert array is not None
         img = PILImage.fromarray(array, mode=mode)
-        return cls.from_pil(img, uri)
+        return cls.from_pil(img, uri, format=format, **kwargs)
 
     @staticmethod
-    def from_pil(img: PILImage, uri: Union[str, Path]) -> Image:
+    def from_pil(
```
eddyxu, line 94: giving that we will support various types from `__init__` , we might be able to consolidate this and `from_array` and etc factory methods together.

Not need to be done here tho.

eddyxu (approved):

PR #179, opened 2021-04-02
Image can be on both inline or external storage

eddyxu:


src/main/scala/org/apache/spark/sql/rikai/Image.scala
```diff
@@ -28,21 +28,37 @@ private[spark] class ImageType extends UserDefinedType[Image] {
...
+        new Image(
+          row.isNullAt(0) match {
+            case true  => None
+            case false => Some(row.getBinary(0))
+          },
+          row.isNullAt(1) match {
+            case true  => None
+            case false => Some(row.getString(1).toString)
+          }
+        )
```
changhiskhan, line 57: this will throw if it's not an InternalRow.

eddyxu: It seems wont happen? this is merely used for doing the type conversion., for example, `spark.ml` does the same

https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/linalg/VectorUDT.scala#L54-L69

changhiskhan: for scala, exhaustive pattern matching is considered best practice. In 2.13.4 for example strict exhaustive pattern matching is on by default. In older versions it's just a warning.

src/main/scala/org/apache/spark/sql/rikai/Image.scala
```diff
@@ -28,21 +28,33 @@ private[spark] class ImageType extends UserDefinedType[Image] {
...
+    obj.uri.map(x => row.update(1, UTF8String.fromString(x)))
     row
   }
 
   override def deserialize(datum: Any): Image =
     datum match {
-      case row: InternalRow => new Image(row.getString(0).toString())
+      case row: InternalRow =>
+        new Image(
+          row.isNullAt(0) match {
```
changhiskhan, line 49: does `Option.unless(row.isNullAt(0))(row.getBinary(0))` and `Option.when(row.isNullAt(1))(row.getString(1).toString)` work?

python/rikai/types/vision.py
```diff
@@ -43,15 +43,20 @@ class Image(ToNumpy, ToPIL, Asset, Displayable):
...
     """
 
     __UDT__ = ImageType()
 
-    def __init__(self, uri: Union[str, Path]):
-        super().__init__(uri)
-        self._cached_data = None
+    def __init__(
+        self,
+        data: Optional[bytes] = None,
```
changhiskhan, line 56: will creating images from embedded data be more common than creating images from uri's? If not maybe swap the order so you don't need to explicitly specify the parameter name?

Another thing to consider might be that while it makes sense for the Asset ABC to separate these parameters, maybe the "user-facing" classes like `Image` can just use a single parameter (non-option) and figure out whether it's the data or the uri based on the input type?

changhiskhan (commented):
good stuff. i think this is a solid base to build from. From this change i think we have lots of room for workflow improvement. Two major things that come to mind are:

1. Provide APIs to externalize/inline whole Image columns at once
2. Create the rikai datasource in spark and be able to read large zip files
(but obviously don't have to be done in this PR)

python/rikai/types/vision.py
eddyxu: good point.  let me update this PR>

src/main/scala/org/apache/spark/sql/rikai/Image.scala
eddyxu: lemme try some scala magic

eddyxu:
i do imagine a transparent user experience could will be sth like:

```
class Image:
  def __init__(self, data):
       if len(data) > rikai.option.get("rikai.asset.embedded_max_size"):
             self.uri = upload_to_external(data, rikai.option.get("rikai.asset.default_dir")
```

So that we could have good tuning on parquet scan / random access performance without user interaction.

src/main/scala/org/apache/spark/sql/rikai/Image.scala
eddyxu: What exception do you suggest to throw here?

src/main/scala/org/apache/spark/sql/rikai/Image.scala
eddyxu: Hmm, in. scala 2.12 `Option` does not have this method

src/main/scala/org/apache/spark/sql/rikai/Image.scala
changhiskhan: no strong opinion here. Maybe IllegalArgumentException?

changhiskhan (approved):

PR #181, opened 2021-04-02
Abstract out ModelSpec

changhiskhan:
1. Make ModelSpec an ABC that can be derived by different types of Registries
2. Change the YAML model version property from number to string (e.g., don't want to deal with floating point precision)

This is a preparatory refactor to adding MLFlow Registry

### DIFF
--- a/python/rikai/spark/sql/codegen/pytorch.py
+++ b/python/rikai/spark/sql/codegen/pytorch.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 
 from pathlib import Path
-from typing import Callable, Dict, Iterator, Optional, Union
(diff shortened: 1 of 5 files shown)


python/rikai/spark/sql/codegen/pytorch.py
```diff
@@ -25,29 +25,20 @@
...
+    spec : ModelSpec
+        the model specifications object
 
     Returns
     -------
     A Spark Pandas UDF.
     """
-    options = {} if options is None else options
+
+    options = {} if spec.options is None else spec.options
```
eddyxu, line 41: This can be handled by `spec` object?

eddyxu (approved):

Issue #25, opened 2021-01-19
Authenticate readthedocs oauth app to Rikai


eddyxu:
So that we can build / test doc for each PR.

PR #183, opened 2021-04-03
[DOC]Add missed dependencies on readthedocs build

eddyxu:


eddyxu:
This is also relevant to #25 , as we dont really know what is the build result for each commit at the time.

changhiskhan (approved):

PR #184, opened 2021-04-03
Fix flaky yaml spec tests

eddyxu:


eddyxu:
Re-triggered the tests 3 times, all passed.

python/rikai/spark/sql/codegen/pytorch.py
```diff
@@ -24,6 +25,9 @@
 from rikai.io import open_uri
 from rikai.torch.pandas import PandasDataset
 
+DEFAULT_NUM_WORKERS = 8
```
changhiskhan, line 28: nit: maybe add this to rikai.options?

changhiskhan (approved):
good stuff

python/rikai/spark/sql/codegen/pytorch.py
eddyxu: will do later

PR #185, opened 2021-04-03
Fix rtd build failure

eddyxu:
Fix pytorch install on readthedoc.io

### DIFF
--- a/python/deps/requirements_rtd.txt
+++ b/python/deps/requirements_rtd.txt
@@ -6,6 +6,6 @@ Pillow
 pyarrow>=2.0
 pycocotools==2.0.2
 pyspark>=3
-pytorch
+torch
 pyyaml
 semver


eddyxu:
Fix https://readthedocs.org/projects/rikai/builds/13415620/

python/deps/requirements_rtd.txt
```diff
@@ -6,6 +6,6 @@ Pillow
 pyarrow>=2.0
 pycocotools==2.0.2
 pyspark>=3
-pytorch
+torch
```
changhiskhan, line 9: is this because it's pytorch in conda install but torch with pip install?

changhiskhan (approved):

python/deps/requirements_rtd.txt
eddyxu: Yes, seems to be so.

PR #191, opened 2021-04-03
Implement MlflowModelSpec and create mlflowmodel spec based on run id

changhiskhan:
Remaining pieces:

1. Scala-side changes
2. Hook up the spec to the registry
3. Some logging utility to track correct tags/parameters during training (e.g., `rikai.mlflow.log_spec(flavor=, schema=, pre_processing=, ...)`)

Later on we may want to integrate with mlflow model registry but it's not super mature yet.


python/tests/spark/sql/codegen/test_mlflow_registry.py
```diff
@@ -0,0 +1,82 @@
...
+                "rikai.transforms.pre": (
+                    "rikai.contrib.torch.transforms."
+                    "fasterrcnn_resnet50_fpn.pre_processing"
+                ),
+                "rikai.transforms.post": (
+                    "rikai.contrib.torch.transforms."
+                    "fasterrcnn_resnet50_fpn."
+                    "post_processing"
+                ),
+                "rikai.model.artifact.uri": resnet_model_uri,
```
eddyxu, line 54: we should support `mlflow.<flavor>.load_func/load_model` from udf?

This file based URI might be just an accident support in mlflow

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,192 @@
...
+    str
+        Spark UDF function name for the generated data.
+    """
+    try:
+        import mlflow
+    except ImportError:
+        raise ImportError(
+            "Couldn't import mlflow. Please make sure to "
+            "`pip install mlflow` explicitly or install "
+            "the correct extras like `pip install rikai[all]`"
```
eddyxu, line 148: `rikai[all]` is so big now, maybe we should just suggest to do `pip install rikai[mlflow]`?

python/tests/spark/sql/codegen/test_mlflow_registry.py
changhiskhan: You mean like "ML_PREDICT(mlflow.pytorch.<model_name>[.latest | .<version>], ...)"? That's a good idea actually.

eddyxu: i meant `https://www.mlflow.org/docs/latest/model-registry.html#fetching-an-mlflow-model-from-the-model-registry` 

The case you mentioned can do too.

Currently `ML_PREDICT` does support in flight model loading like

```
ML_PREDICT("mlflow://models/model_name", ...) 
```

and it will load a temporary model, which is not registered into ModelCatalog

python/tests/spark/sql/codegen/test_mlflow_registry.py
```diff
@@ -0,0 +1,86 @@
...
+def test_mlflow_model_from_runid(spark: SparkSession, mlflow_client):
+    run_id = mlflow_client.search_model_versions("name='test'")[0].run_id
+    spark.sql("CREATE MODEL resnet_m_foo USING 'mlflow://{}'".format(run_id))
+    check_ml_predict(spark, 'resnet_m_foo')
+
+
+@pytest.mark.timeout(60)
+def test_mlflow_model_from_model_version(spark: SparkSession, mlflow_client):
+    spark.sql("CREATE MODEL resnet_m_bar USING 'mlflow://test/1'")
+    check_ml_predict(spark, 'resnet_m_bar')
```
changhiskhan, line 86: both these tests fail with the following:
```
E                   py4j.protocol.Py4JJavaError: An error occurred while calling o40.sql.
E                   : ai.eto.rikai.sql.model.ModelRegistryAlreadyExistException: ModelRegistry(file) exists
E                   	at ai.eto.rikai.sql.model.Registry$.$anonfun$registerAll$2(Registry.scala:82)
E                   	at ai.eto.rikai.sql.model.Registry$.$anonfun$registerAll$2$adapted(Registry.scala:69)
E                   	at scala.collection.TraversableLike$WithFilter.$anonfun$foreach$1(TraversableLike.scala:877)
E                   	at scala.collection.immutable.HashMap$HashMap1.foreach(HashMap.scala:234)
E                   	at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:468)
E                   	at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:468)
E                   	at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:876)
E                   	at ai.eto.rikai.sql.model.Registry$.registerAll(Registry.scala:69)
E                   	at ai.eto.rikai.sql.spark.parser.RikaiExtSqlParser.registyInitialized$lzycompute(RikaiExtSparkSQLParser.scala:54)
E                   	at ai.eto.rikai.sql.spark.parser.RikaiExtSqlParser.registyInitialized(RikaiExtSparkSQLParser.scala:53)
E                   	at ai.eto.rikai.sql.spark.parser.RikaiExtSqlParser.parsePlan(RikaiExtSparkSQLParser.scala:58)
...
...
```

@eddyxu i don't see where it should be registering multiple times...

changhiskhan: nvm something was screwed up in my local env.

python/tests/spark/sql/codegen/test_mlflow_registry.py
changhiskhan: oh i see yeah i added that already.

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: sg changed.

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,275 @@
...
+class MlflowModelSpec(ModelSpec):
+    """Model Spec.
+
+    Parameters
+    ----------
+    run : mlflow.entities.Run
+        the Run object containing the spec info and ref to model artifact
+    options : Dict[str, Any], optional
+        Additionally options. If the same option exists in spec already,
+        it will be overridden.
```
eddyxu, line 52: add doc for all parameters?

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,275 @@
...
+
+    @property
+    def end_time(self):
+        return self._run.info.end_time
+
+    @property
+    def artifact(self) -> Any:
+        """Return Model artifact"""
+        if self._artifact is None:
+            mlflow.set_tracking_uri(self.tracking_uri)
```
eddyxu, line 89: `self.tracking_uri` can be None?  can `mlflow.set_tracking_uri` take it?

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,275 @@
...
+            self._artifact = getattr(mlflow, self.flavor).load_model(self.uri)
```
eddyxu, line 90: does this mean that model will be picked via the content and send to workers?

I can see models that are memory bound (i.e., more than 10GB), will that be problematic? Also, this might means that the same model is in JVM, python interpreter as well as in GPU.

Why not call `load_model` in the `pandas_udf` itself tho? in that case, it does not take any spark/jvm seen memory.   Also this might be dangerous to load a bunch of models at once?

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,275 @@
...
+    model: Any,
+    artifact_path: str,
+    flavor: str,
+    schema: str,
+    pre_processing: Optional[str] = None,
+    post_processing: Optional[str] = None,
+    registered_model_name: Optional[str] = None,
+    **kwargs,
+):
+    """Convenience function to log the model with information needed by rikai
```
eddyxu, line 183: Is this used in training process

could be more explicitly by offering examples?

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,275 @@
...
+        self._spark = spark
+        self._jvm = spark.sparkContext._jvm
+
+    def __repr__(self):
+        return "MlflowRegistry"
+
+    def resolve(self, uri: str, name: str, options: Dict[str, str]):
+        logger.info(f"Resolving model {name} from {uri}")
+        tracking_uri = self._spark.conf.get(
+            "rikai.sql.ml.registry.mlflow.tracking_uri"
```
eddyxu, line 236: CONST in `rikai.conf`?

python/rikai/spark/sql/codegen/pytorch.py
```diff
@@ -41,20 +41,23 @@ def generate_udf(spec: "rikai.spark.sql.codegen.base.ModelSpec"):
...
+    batch_size = int(spec.options.get("batch_size", DEFAULT_BATCH_SIZE))
 
     def torch_inference_udf(
         iter: Iterator[pd.DataFrame],
     ) -> Iterator[pd.DataFrame]:
 
-        with open_uri(spec.uri) as fobj:
-            model = torch.load(fobj)
+        if spec.artifact:
+            model = spec.artifact
```
eddyxu, line 57: we should consolidate these two methods.

python/tests/spark/sql/codegen/test_mlflow_registry.py
```diff
@@ -0,0 +1,104 @@
...
+            "fasterrcnn_resnet50_fpn.pre_processing"
+        )
+        post_processing = (
+            "rikai.contrib.torch.transforms."
+            "fasterrcnn_resnet50_fpn.post_processing"
+        )
+        log_model(
+            model,
+            artifact_path,
+            "pytorch",
```
eddyxu, line 60: does this mean that it can only support the mlflow models where these tags exist?

Do we want to support vanilla mlflow models?

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: Shouldn't be none. i can add an assert

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: yeah, potentially. Need to make sure to get the tracking_uri to the worker too is all.

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: yes during training. yeah i'll add an example.

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: good idea

python/rikai/spark/sql/codegen/pytorch.py
changhiskhan: the mlflow one isn't just a simple file. it's Mlflow Model format with pickle information as well.

python/tests/spark/sql/codegen/test_mlflow_registry.py
changhiskhan: if we do then we'd have to pass that extra information into CREATE MODEL or ML_PREDICT somehow. We can add support for that later on i just figured version 1 we keep it simple

python/rikai/spark/sql/codegen/pytorch.py
eddyxu: yea, what i meant is that we can deligate the "load model" impl behind "ModelSpec"

For example: `FileSystemModelSpec.load_model()` and `MLflowModelSpec.load_model()`, so that we dont need `if/else` here.

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: added. update coming soon

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: made it a non-optional argument

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: switched to load_model instead

python/rikai/contrib/torch/transforms/fasterrcnn_resnet50_fpn.py
```diff
@@ -54,3 +54,9 @@ def post_process_func(batch):
         return results
 
     return post_process_func
+
+
+OUTPUT_SCHEMA = (
```
eddyxu, line 59: was thinking to standardize how to write pre/post/schema combined.  might have a RFC later.

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,303 @@
...
+    def start_time(self):
+        return self._run.info.start_time
+
+    @property
+    def end_time(self):
+        return self._run.info.end_time
+
+    def load_model(self) -> Any:
+        """Load the model artifact specified in this spec"""
+        old_uri = mlflow.get_tracking_uri()
```
eddyxu, line 90: why would `old_uri` be different from `self.tracking_uri`? mind to add some comments here?

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def _run_to_spec_dict(self):
+        """Convert the Run into a ModelSpec
+
+        Returns
+        -------
+        spec: Dict[str, Any]
+        """
+        tags = self._run.data.tags
+        spec = {
```
eddyxu, line 105: do we expect these keys do not exist? if so, lets throw some actionable errors?

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,303 @@
...
+    Returns
+    -------
+    str
+        Spark UDF function name for the generated data.
+    """
+    try:
+        spec = MlflowModelSpec(run, tracking_uri, options=options)
+    except Exception as e:
+        to_spec_msg = (
+            "Could not create well-formed ModelSpec from run {}."
```
eddyxu, line 171: do we expect this is the only exception (spec formation?) should we differentiate exceptions?

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,303 @@
...
+        Full python module path of the post-processing transforms
+    registered_model_name: str, default None
+        Model name in the mlflow model registry
+    kwargs: dict
+        Passed to `mlflow.<flavor>.log_model`
+
+    Examples
+    --------
+    # import mlflow.pytorch  <= vanilla mlflow usage
+    from rikai.spark.sql.codegen.mlflow_registry import log_model
```
eddyxu, line 215: we might want to move this to `rikai.mlflow` package if we expect users to use this API?  this `rikai.spark.sql.codegen` is long and sounds like internal package.

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,303 @@
...
+    with mlflow.start_run() as run:
+
+        # Training loop
+        # ...
+
+        # Assume `model` is the trained model from the training loop
+        log_model(model, "model", "pytorch",
+                  schema=rikai.contrib.torch.transforms.fasterrcnn_resnet50_fpn.OUTPUT_SCHEMA,
+                  pre_processing="rikai.contrib.torch.transforms.fasterrcnn_resnet50_fpn.pre_processing",
+                  post_processing="rikai.contrib.torch.transforms.fasterrcnn_resnet50_fpn.post_processing",
```
eddyxu, line 227: A little worry that users will use the actual function here. 

We might want to reconsider the way to write spec.

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,303 @@
...
+        self._spark = spark
+        self._jvm = spark.sparkContext._jvm
+
+    def __repr__(self):
+        return "MlflowRegistry"
+
+    def resolve(self, uri: str, name: str, options: Dict[str, str]):
+        logger.info(f"Resolving model {name} from {uri}")
+        tracking_uri = self._spark.conf.get(CONF_MLFLOW_TRACKING_URI)
+        client = MlflowClient(tracking_uri)
```
eddyxu, line 266: should we keep a `self._client`? `resolve` will be called everytime a `CREATE MODEL` is called.

src/main/scala/ai/eto/rikai/sql/model/mlflow/MlflowModel.scala
```diff
@@ -0,0 +1,51 @@
...
+    val name: String,
+    val uri: String,
+    val funcName: String
+) extends Model
+    with SparkRunnable {
+
+  override def toString: String = s"MlflowModel(name=${name}, uri=${uri})"
+
+  /** Convert a [[Model]] to a Spark Expression in Spark SQL's logical plan. */
+  override def asSpark(args: Seq[Expression]): Expression = {
```
eddyxu, line 44: nit , should we pull this up as a "SparkUDFModel"?

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: it's not heavy weight but sure I can do that. Probably need to check if the tracking uri changes and then re-instantiate the client

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: The backup is we can ask the users to supply the last 4 rikai specific arguments in the OPTIONS clause of CREATE MODEL. But honestly that's much more complicated for the SQL ML user.

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: yeah, that's not a bad idea

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: what did you have in mind?

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: just being defensive

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: yeah i can add a message about they should use rikai.mlflow.log_model, or add these to the options.

python/rikai/contrib/torch/transforms/fasterrcnn_resnet50_fpn.py
changhiskhan: would be much easier in a statically typed language lol

python/rikai/spark/sql/codegen/mlflow_registry.py
eddyxu: for example, is it possible to make connection in `__init__` and raises NetworkException / PermissionDenied or some sort.

changhiskhan: for now i'll just let individual exceptions bubble up?

src/main/scala/ai/eto/rikai/sql/model/mlflow/MlflowModel.scala
changhiskhan: done

python/rikai/contrib/torch/transforms/fasterrcnn_resnet50_fpn.py
eddyxu: lol we can enforce type hints for pre/post processing? 
to some extend, the output schema here is the return type of `post_processing` routine.

python/rikai/spark/sql/codegen/mlflow_logger.py
```diff
@@ -0,0 +1,123 @@
...
+import mlflow
+
+from rikai.conf import (
+    CONF_MLFLOW_ARTIFACT_PATH,
+    CONF_MLFLOW_MODEL_FLAVOR,
+    CONF_MLFLOW_OUTPUT_SCHEMA,
+    CONF_MLFLOW_POST_PROCESSING,
+    CONF_MLFLOW_PRE_PROCESSING,
+    CONF_MLFLOW_SPEC_VERSION,
+    CONF_MLFLOW_TRACKING_URI,
```
eddyxu, line 28: nit: these keys are supposed to be used in `mlflow logger`? a bit confused that they are in the same module with `rikai.config.option` as well as the same configure key namespace `rikai.*`.

dont need to change anything tho.

python/rikai/spark/sql/codegen/mlflow_logger.py
```diff
@@ -0,0 +1,123 @@
...
+        # Log PyTorch model
+        with mlflow.start_run() as run:
+
+            # Training loop
+            # ...
+
+            # Assume `model` is the trained model from the training loop
+            rikai.mlflow.pytorch.log_model(model, "model",
+                      schema=rikai.contrib.torch.transforms.fasterrcnn_resnet50_fpn.OUTPUT_SCHEMA,
+                      pre_processing="rikai.contrib.torch.transforms.fasterrcnn_resnet50_fpn.pre_processing",
```
eddyxu, line 97: is it possible that we take this parameter as a function instead of a string?

a string could make refactory much harder.

python/rikai/spark/sql/codegen/mlflow_logger.py
```diff
@@ -0,0 +1,123 @@
...
+    CONF_MLFLOW_PRE_PROCESSING,
+    CONF_MLFLOW_SPEC_VERSION,
+    CONF_MLFLOW_TRACKING_URI,
+)
+
+KNOWN_FLAVORS = ["pytorch"]
+__all__ = KNOWN_FLAVORS
+
+
+class MlflowLogger:
```
eddyxu, line 35: so this logger is still in `codegen` module?

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -0,0 +1,307 @@
...
+        the model spec info dict
+    name : str
+        The name of the model in the catalog
+
+    Returns
+    -------
+    str
+        Spark UDF function name for the generated data.
+    """
+    udf = udf_from_spec(spec)
```
eddyxu, line 198: is this method exact the same as in `FSRegistry`?

python/rikai/spark/sql/codegen/fs.py
```diff
@@ -99,13 +108,12 @@ def __repr__(self):
...
     def resolve(self, uri: str, name: str, options: Dict[str, str]):
         logger.info(f"Resolving model {name} from {uri}")
-
         if uri.endswith(".yml") or uri.endswith(".yaml"):
             func_name = codegen_from_yaml(self._spark, uri, name, options)
         else:
             raise ValueError(f"Model URI is not supported: {uri}")
 
-        model = self._jvm.ai.eto.rikai.sql.model.fs.FileSystemModel(
+        model = self._jvm.ai.eto.rikai.sql.model.SparkUDFModel(
```
eddyxu, line 116: 👏

eddyxu (approved):
small nits. LGTM overall

python/rikai/spark/sql/codegen/mlflow_logger.py
changhiskhan: It's imported as rikai.mlflow in rikia.__init__.py (see test case)

python/rikai/spark/sql/codegen/mlflow_logger.py
changhiskhan: so turn the function into a string inside log_model then write it to the server?

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: `udf_from_spec` is a shared function (comes from base.py)
FileSystemRegistry reads from the spec file to generate the spec. MlflowRegistry will get attributes from tracking server to generate the spec dict.

python/rikai/spark/sql/codegen/mlflow_logger.py
changhiskhan: i'll move them to mlflow_logger for now (they're not registered in rikai.config.options right now)

PR #203, opened 2021-04-08
Load image from bytes

smellslikeml:
Updates `rikai.spark.functions.image` to load from bytes reflecting recent changes to `ImageType`.


python/rikai/spark/functions/vision.py
```diff
@@ -47,9 +49,17 @@
 
 
 @udf(returnType=ImageType())
-def image(uri: str) -> Image:
-    """Build an :py:class:`Image` from a URI."""
-    return Image(uri)
+def image(image: Union[bytes, bytearray, IOBase, str, Path]) -> Image:
```
changhiskhan, line 52: 1. i would call the function `to_image` instead?
2. the input argument call it `image_data` or something?
3. what produces the IOBase / bytes / bytearray ? does reading from Spark image data source produce all 3?

python/rikai/spark/functions/vision.py
```diff
@@ -47,9 +49,17 @@
...
+    """Build an :py:class:`Image` from
+    bytes, file-like object, str, or :py:class:`~pathlib.Path`."""
+    data, uri = None, None
+    if isinstance(image, IOBase):
```
changhiskhan, line 56: can you add 3 unit tests for each of the 3 branches in this if-else here?

changhiskhan (commented):
very simple and useful. cool!

oh btw, can you also change `video_to_frames` so that it creates embedded Images instead externalized ones? That should simplify the code/workflow for video analysis. (doesn't have to be in this PR)

python/rikai/spark/functions/vision.py
smellslikeml: updated the naming conventions, modeled the input types from [here](https://github.com/eto-ai/rikai/blob/4a97e9d9cf4579ceae9efa90332affaab2714430/python/rikai/types/vision.py#L55) but Spark Images keeps `bytes` in the `Image.data` column.

As VideoCapture extracts numpy arrays, I can update `video_to_frames` with this [PR](https://github.com/smellslikeml/rikai/tree/ffmpeg-load-video) to avoid decompression.

python/rikai/spark/functions/vision.py
smellslikeml: got it but I eliminated `IOBase` since we don't have a spark column of this type.

changhiskhan (approved):
lgtm thanks!
Before you merge it, can you add an example to the udf docstring (see section 15. Examples https://numpydoc.readthedocs.io/en/latest/format.html)

python/rikai/spark/functions/vision.py
```diff
@@ -47,9 +48,28 @@
...
+
+    Parameters
+    ----------
+    image_data : bytes, bytearray, str, Path
+        The resource identifier or bytes of the source image.
+
+    Return
+    ------
+    Image
+        Return an Image pointed to the new URI
```
changhiskhan, line 63:
```
Returns
--------
img: Image
    An Image from the given embedded data or URI
```

changhiskhan (approved):
i re-ran the tests and they passed. Thanks for addressing the comments. Just one last nit on the doc please feel free to merge

PR #204, opened 2021-04-08
Return passthrough func if pre/post processing transforms are not set

eddyxu:


python/tests/spark/sql/codegen/test_fs.py
```diff
@@ -164,14 +160,13 @@ def test_yaml_model(spark: SparkSession, resnet_spec: str):
...
-                            "boxes",
-                            ArrayType(ArrayType(FloatType())),
+                            "boxes", ArrayType(ArrayType(FloatType()))
                         ),
                         StructField("scores", ArrayType(FloatType())),
                         StructField("labels", ArrayType(IntegerType())),
                     ]
                 ),
-            ),
+            )
```
eddyxu, line 169: these were auto-formatted by `black`. i can revert them if desired.

changhiskhan (approved):

PR #210, opened 2021-04-11
mlflow registry docs and updates

changhiskhan:
1. Updated README and sqlml docs for mlflow registry integration
2. Add unit test for CREATE MODEL on plain vanilla mlflow model (without the custom logger)
3. To make #2 work, pass in the options dict to model resolve

I think it might make sense to add a "MODELPROPERTIES" or similar clause so that we can separate the resolve time options vs run time options.

### DIFF
--- a/docs/sqlml.rst
+++ b/docs/sqlml.rst
@@ -75,10 +75,10 @@ Rikai extends Spark SQL with four more SQL statements:
         { DESC | DESCRIBE } MODEL model_name;
 
         # Show all models
-        SHOW MODELS
+        SHOW MODELS;
 
         # Delete a model
-        DROP MODEL model_name
+        DROP MODEL model_name;
 
 Rikai uses URL schema to decide which Model Registry to be used to resolve a
 ML Model. Once one ML model is via ``CREATE MODEL``,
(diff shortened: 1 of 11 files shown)


src/main/scala/ai/eto/rikai/sql/model/fs/FileSystemRegistry.scala
```diff
@@ -41,8 +41,12 @@ class FileSystemRegistry(val conf: Map[String, String])
...
   @throws[ModelNotFoundException]
-  override def resolve(uri: String, name: Option[String]): Model = {
+  override def resolve(
+      uri: String,
+      name: Option[String],
+      options: Option[Map[String, String]]
+  ): Model = {
     logger.info(s"Resolving ML model from ${uri}")
-    Python.resolve(pyClass, uri, name, Map.empty)
+    Python.resolve(pyClass, uri, name, options.getOrElse(Map.empty))
```
eddyxu, line 50: Do you think we should actually pass a `ModelSpec` from jvm to python, merged in python, and return a Model back?

changhiskhan: Is this because you're thinking about reading the ModelSpec out the catalog db and then passing it to python to be "merged" and turned into an actual Model instance? or this is because of some other reason?

eddyxu: I see that we will have more parameters than these three in the future. Using an interface ModelSpec here can make the code more stable

docs/sqlml.rst
```diff
@@ -39,6 +39,10 @@ Before we can use ``Rikai SQL-ML``, we need to configure SparkSession:
             "rikai.sql.ml.registry.file.impl",
             "ai.eto.rikai.sql.model.fs.FileSystemRegistry",
         )
+        .config(
```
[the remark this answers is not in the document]
changhiskhan: not a bad idea. And if users specify an alternate then the one in the spark config will override.

src/main/scala/ai/eto/rikai/sql/model/fs/FileSystemRegistry.scala
eddyxu: we dont need to change this now.

eddyxu (approved):
LGTM

Issue #212, opened 2021-04-12
Support qualified name as keys in CREATE MODEL options


eddyxu:
We can support qualified name "foo.bar.zoo" in `CREATE MODEL OPTIONS`

eddyxu:
Fixed in #223

PR #219, opened 2021-04-12
Include generated module and also non-python files

changhiskhan:
also add STRING as valid optionkey

### DIFF
--- a/python/rikai/spark/sql/generated/__init__.py
+++ b/python/rikai/spark/sql/generated/__init__.py
@@ -1,0 +1,13 @@
+#  Copyright (c) 2021 Rikai Authors
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
--- a/Makefile
+++ b/Makefile
@@ -18,12 +18,14 @@ all: antlr
 antlr: python/rikai/spark/sql/generated/RikaiModelSchemaParser.py
 .PHONY: antlr
 
+# On ubuntu apt installs only antlr4 so create symlink like (sudo ln -sf /usr/bin/antlr4 /usr/local/bin/antlr)
 python/rikai/spark/sql/generated/RikaiModelSchemaParser.py: src/main/antlr4/org/apache/spark/sql/ml/parser/RikaiModelSchema.g4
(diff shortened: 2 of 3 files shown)


python/rikai/spark/sql/generated/__init__.py
```diff
@@ -0,0 +1,13 @@
+#  Copyright (c) 2021 Rikai Authors
```
eddyxu, line 1: will this file be deleted after a new `make generated`?

src/main/antlr4/org/apache/spark/sql/ml/parser/RikaiExtSqlBase.g4
```diff
@@ -84,6 +84,7 @@ option
 
 optionKey
     : identifier
+    | STRING
```
eddyxu, line 87: is this relevant?

changhiskhan: limits you to variable names only. e.g., you can't specify a spark config variable name as the option key (e.g., "rikai.model.config")

python/rikai/spark/sql/generated/__init__.py
changhiskhan: i added touch in the makefile

src/main/antlr4/org/apache/spark/sql/ml/parser/RikaiExtSqlBase.g4
eddyxu: as key, it should be an quanlifiedname , as `OPTIONS (rikai.model.config = 123)` but not a STRING `OPTIONS ("rikai.model.config" = 123)`?

eddyxu: Filed #212

python/rikai/spark/sql/generated/__init__.py
eddyxu: nit, i am confused whether this file will be regererated via `make antrl`.
If yes, then the content in this file will be lost next time we run `make antlr`.
If no, we dont need to `touch` in Makefile?

eddyxu (approved):
LGTM.  Left a question , but it does not block merging this PR

python/rikai/spark/sql/generated/__init__.py
changhiskhan: Locally it depends on whether you deleted the whole directory contents or just removed one file. If you just change the g4 file without touching the generated directory make will do nothing (at least on Ubuntu). So we can't really guarantee one way or the other but this at least ensures we don't run into the same situation where those python files don't get added to the wheel. Touch will do nothing if the file already exists so it's not an issue either.

changhiskhan: This file is not by default generated by the antlr command

changhiskhan: Init will have no contents other than the license. It's just to make that it's a python module to setup

PR #220, opened 2021-04-13
Support relative path as model uri.

eddyxu:


python/rikai/spark/sql/codegen/fs.py
```diff
@@ -90,8 +103,7 @@ def codegen_from_yaml(
     str
         Spark UDF function name for the generated data.
     """
-    with open_uri(uri) as fobj:
-        spec = FileModelSpec(fobj, options=options)
+    spec = FileModelSpec(uri, options=options)
```
changhiskhan, line 106: s/uri/spec_uri/g for clarity?

changhiskhan (approved):

PR #225, opened 2021-04-13
Add support for negative numbers

changhiskhan:
Change DECIMAL_VALUE -> FLOATING_VALUE. We should use DECIMAL to indicate number
base if we need to support binary/hex/oct numbers in the future.


eddyxu (approved):
LGTM

PR #230, opened 2021-04-14
A few minor build fixes

changhiskhan:
1. Set SPARK_LOCAL_IP and increase spark.port.maxRetries to hopefully fix test flakiness for Spark
2. Automatically delete the rikai ivy cache when publishLocal.

publishLocal sends the jar to ivy local repository. When we run tests which refers to
the rikai jar as a spark package, it loads the local jar and caches it. This then makes it
so that subsequent publishLocal is shadowed by the cached jar unless it is manually removed.
Instead we override publishLocal task to first delete the ivyHome/cache/ai.eto/rikai_2.12 directory

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -44,6 +44,8 @@ jobs:
       working-directory: python
       run: |
         pytest -v --durations=10 --ignore=tests/parquet/internal
+      env:
+        SPARK_LOCAL_IP: "127.0.0.1"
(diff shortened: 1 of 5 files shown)


eddyxu:
>  When we run tests which refers to the rikai jar as a spark package, it loads the local jar and caches it.

Is this for local development or GH action?

eddyxu (approved):

PR #232, opened 2021-04-15
Allow user to set default model uri root

changhiskhan:
Since FileSystemRegistry supports multiple schemes, having a default
registry implementation with only relative paths for model uri is
ambiguous. Instead, we can support setting a default model uri root.
This can be file:/ or mlflow:/, or s3://bucket/ gs://bucket/.


python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -224,74 +224,82 @@ def resolve(self, spec):
...
-        raise SpecError(
-            "No versions found for model {} in stage {}".format(model, stage)
+    def is_model(self, entity_reference):
+        """
+        Returns True is the given mlflow entity reference is a registered model
+        """
+        try:
+            self.tracking_client.get_registered_model(entity_reference)
+            return True
+        except Exception:
```
eddyxu, line 255: a `SocketTimeout` will return False as well. Is it desired?

python/tests/conftest.py
```diff
@@ -38,6 +38,7 @@ def spark(tmp_path_factory) -> SparkSession:
             "spark.sql.extensions",
             "ai.eto.rikai.sql.spark.RikaiSparkSessionExtensions",
         )
+        .config("rikai.sql.ml.registry.uri.root", "mlflow:/")
```
eddyxu, line 41: are we going to make `mlflow` the main test case?

src/main/scala/ai/eto/rikai/sql/model/Registry.scala
```diff
@@ -42,9 +44,8 @@ private[rikai] object Registry {
 
   val REGISTRY_IMPL_PREFIX = "rikai.sql.ml.registry."
   val REGISTRY_IMPL_SUFFIX = ".impl"
-  val REGISTRY_IMPL_DEFAULT_KEY = "rikai.sql.ml.registry.impl"
+  val DEFAULT_URI_ROOT_KEY = "rikai.sql.ml.registry.uri.root"
```
eddyxu, line 47: mind to write some doc here THAT IS not only scheme?

`uri.root` is not an concept w/o confusions.

python/tests/conftest.py
changhiskhan: what do you think would be the best way to make this limited in scope? this config variable is used once in Registry.registerAll and then that's it.

python/rikai/spark/sql/codegen/mlflow_registry.py
changhiskhan: THis is  trying to figure out whether the input is a model name or a run id. In the case of a sockettimeout it would think it's a runid and try to do get_run(entity_reference) which would also fail. 

Three alternatives:

have mlflow+models:/<model_name>/<version>
or mlflow:/models/<model_name>/<version>
or just not support direct references to a runid (only support mlflow model registry references)

src/main/scala/ai/eto/rikai/sql/model/Registry.scala
changhiskhan: fair. will add documentation. but first, sleep

changhiskhan: ok docs updated.

python/tests/conftest.py
eddyxu: can we provide a `conftest.py`  for `rikai/python/tests/mlflow`?

python/rikai/spark/sql/codegen/mlflow_registry.py
eddyxu: can we register `MLFlowRegistery` with both `rikai.ml.sql.registry.models.impl` AND  `rikai.ml.sql.registry.runs.impl` ?

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -224,74 +224,82 @@ def resolve(self, spec):
         uri = spec.getUri()
         options = spec.getOptions()
         logger.info(f"Resolving model {name} from {uri}")
-        model_uri, tags, params = _get_model_info(uri, self.tracking_client)
+        parsed = urlparse(uri)
+        if not parsed.scheme:
+            raise ValueError("Scheme must be mlflow. How did you get here?")
+        parts = parsed.path.strip("/").split("/", 1)
```
eddyxu, line 230: maybe this is unavoid-able, these URI stuff are so complicated now..

python/rikai/spark/sql/codegen/mlflow_registry.py
eddyxu: i guess my main desire here is that users can use `rikai` in a way that they dont need to convert different URI pattern in mind, for the sake of better UX / `Principle of least astonishment` ;)

changhiskhan: Yeah, I thought about that as well. Very reasonable points.

I decided against surfacing the mlflow:/ and runs:/ schemes to Rikai users because:
1. models:/ and runs:/ make sense in the context of mlflow but are ambiguous in wider contexts.
2. I see the mlflow style "models:/model" and "runs:/run_id" as implementation details internal to mlflow. 
3. I want our users thinking in terms of model names, versions, and at most run_id (only when absolutely necessary) 
4. Having both models:/ and runs:/ means you can have them both as the default uri root.
5. It's likely that SQL ML users won't really be all that familiar with the implementation details of Mlflow so they probably don't even know the "models:/" and "runs:/" URI formats.

python/tests/conftest.py
changhiskhan: was hoping to avoid having multiple Spark sessions in the same test suite. but maybe unavoidable.
i guess the only alternative is adding some reset SQL command to clear the registryMap

python/rikai/spark/sql/codegen/mlflow_registry.py
eddyxu: makes sense. Guess we can see user feedbacks on this later.  LGTM overall.

python/tests/conftest.py
eddyxu: oh, what i suggested that use a separate spark session that is dedicated for `mlflow`

eddyxu (approved):
I think we had good discussion to explore different design choices.  In general, i am ok with the current implement. Would love to see further user feedbacks to decide any adjustment

changhiskhan:
> I think we had good discussion to explore different design choices. In general, i am ok with the current implement. Would love to see further user feedbacks to decide any adjustment

maybe one way to simplify this is to just NOT support referencing bare runid's? And we can put this back in if users are all like "we don't use the model registry but still want to use rikai".

src/main/antlr4/org/apache/spark/sql/ml/parser/RikaiExtSqlBase.g4
```diff
@@ -67,7 +67,7 @@ PREPROCESSOR: 'PREPROCESSOR';
 REPLACE: 'REPLACE';
 RETURNS: 'RETURNS';
 SHOW: 'SHOW';
-STRUCT: 'STRUCT';
+STRUCT: S T R U C T;
```
eddyxu, line 70: Why this change?/

eddyxu: if this is for case insensibility , it will be handled by 

https://github.com/eto-ai/rikai/blob/master/src/main/scala/ai/eto/rikai/sql/spark/parser/RikaiExtSparkSQLParser.scala#L149-L177

python/rikai/spark/sql/generated/RikaiModelSchemaParser.py
```diff
@@ -1,13 +1,9 @@
-# Generated from src/main/antlr4/org/apache/spark/sql/ml/parser/RikaiModelSchema.g4 by ANTLR 4.9.1
+# Generated from src/main/antlr4/org/apache/spark/sql/ml/parser/RikaiModelSchema.g4 by ANTLR 4.7.2
```
eddyxu, line 1: this is a very old antlr version :)

changhiskhan: apt installed :)

src/main/antlr4/org/apache/spark/sql/ml/parser/RikaiExtSqlBase.g4
changhiskhan: make it case insensitive. Scala and Python had mismatching cases.

python/rikai/spark/sql/schema.py
```diff
@@ -115,3 +116,18 @@ def parse_schema(schema_str: str) -> DataType:
     if schema is None:
         raise SchemaError(f"Invalid schema: '{schema_str}'")
     return schema
+
+
+class CaseChangingStream:
```
eddyxu, line 121: is this needed if we dont handle cases in this PR?

eddyxu (approved):
LGTM

PR #235, opened 2021-04-15
Minor fix to load coco dataset

eddyxu:
Fix coco dataset conversion and consistency of print Image in jupyter

### DIFF
--- a/src/main/scala/org/apache/spark/sql/rikai/Image.scala
+++ b/src/main/scala/org/apache/spark/sql/rikai/Image.scala
@@ -70,5 +70,5 @@ private[spark] class ImageType extends UserDefinedType[Image] {
   */
 @SQLUserDefinedType(udt = classOf[ImageType])
 class Image(val data: Option[Array[Byte]], val uri: Option[String]) {
-  override def toString: String = s"Image(uri='$uri')"
+  override def toString: String = s"Image(uri='${uri.getOrElse("<embedded>")}')"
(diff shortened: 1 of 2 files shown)


src/main/scala/org/apache/spark/sql/rikai/Image.scala
```diff
@@ -70,5 +70,5 @@ private[spark] class ImageType extends UserDefinedType[Image] {
   */
 @SQLUserDefinedType(udt = classOf[ImageType])
 class Image(val data: Option[Array[Byte]], val uri: Option[String]) {
-  override def toString: String = s"Image(uri='$uri')"
+  override def toString: String = s"Image(uri='${uri.getOrElse("<embedded>")}')"
```
changhiskhan, line 73: should this be "Image(data=<byte array>)" rather than "Image(uri=<embedded>)" for the embedded case?

changhiskhan (approved):
One minor comment about toString.

As a an aside, in the future we could consider supporting multiple embedded encodings (e.g., base64) and the output here could be the base64 string (which would help debugging).

src/main/scala/org/apache/spark/sql/rikai/Image.scala
eddyxu: hmm, might be simpler just to do `Image({uri | "<embedded>"})` for now? wdyt

PR #238, opened 2021-04-16
rikai: based on ParquetDataSourceV2

da-tubi:


python/tests/parquet/test_dataset.py
```diff
@@ -30,11 +31,14 @@ def test_select_columns(spark: SparkSession, tmp_path: Path):
...
 
-    dataset = Dataset(str(tmp_path), columns=["id", "col1"])
+    dataset = Dataset(rikai_path, columns=["id", "col1"])
     actual = sorted(list(dataset), key=lambda x: x["id"])
 
     assert_count_equal(
         [{"id": 1, "col1": "value"}, {"id": 2, "col1": "more"}], actual
     )
+
+    shutil.rmtree(rikai_path)
```
eddyxu, line 44: `tmp_path` will be automatically deleted by `pytest` framework. we dont need to manually `rmtree`.

https://docs.pytest.org/en/stable/reference.html#std-fixture-tmp_path

src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
```diff
@@ -1 +1 @@
-ai.eto.rikai.RikaiRelationProvider
\ No newline at end of file
+ai.eto.rikai.RikaiDataSourceV2
```
eddyxu, line 1: why V2 :) we dont have V1 yet :p

eddyxu (commented):
i think in the beginning, we'd like to support features similar to delta lakes, i.e., version, delta update, rollback and etc. That's why we did not directly use parquet.

eddyxu:
We can probably to manipulate over the path send to parquet datasource to implement the versions tho.

src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
sadhen: See https://github.com/apache/spark/blob/master/sql/core/src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister

Because it is based on Data Source API V2

sadhen:
The difference between this pr and the master is that this PR are using Data Source API v2 of Apache Spark.

See https://databricks.com/session/apache-spark-data-source-v2

eddyxu:
We need to build a few things to improve training performance.

Rikai format should be a combination of 
* Petastorm https://github.com/uber/petastorm
* Delta lake https://delta.io/

sadhen:
For further customization, we could define:
+ `RikaiFileFormat` like https://github.com/delta-io/delta/blob/master/src/main/scala/org/apache/spark/sql/delta/DeltaFileFormat.scala
+ `RikaiTable` like https://github.com/delta-io/delta/blob/master/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala
+ `RikaiDataSource` like https://github.com/delta-io/delta/blob/master/src/main/scala/org/apache/spark/sql/delta/sources/DeltaDataSource.scala
+ ...

For now, we should keep the implementation as simple as possible.

sadhen:
Here is how CarbonData defines the file format:

https://github.com/apache/carbondata/blob/master/integration/spark/src/main/scala/org/apache/spark/sql/carbondata/execution/datasources/SparkCarbonFileFormat.scala

sadhen:
CarbonData are using Apache Spark 2.x. For Apache Spark 3.x, we need to provide a RikaiWriteBuilder like https://github.com/apache/spark/blob/v3.1.1/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/parquet/ParquetWriteBuilder.scala

sadhen:
@changhiskhan  What's your opinion? Two option:

1. Merge this PR and use `spark.read.format("rikai")` and `df.write.format("rikia")`
2. Close this PR and use `spark.read.format("parquet")` and `df.write.format("parquet")`

changhiskhan:
@da-tubi do we want to keep this open?

PR #240, opened 2021-04-17
Use relative scale of bounding box in ground truth and detections.

eddyxu:
Use relative scale of bound box in both ground truth and detections, so that it is easy to evaluate the performance without scaling again.


eddyxu:
coco bbox is scaled to the original image, while each model will probably resize to some other sizes, using the relative position `[0, 1]` can make the bbox comparable directly from ground truth to detection output

smellslikeml (approved):
LGTM

PR #243, opened 2021-04-17
Support cropping image using Image and Box2d

eddyxu:
Convient methods to crop images into smaller patches.


python/rikai/types/vision.py
```diff
@@ -67,9 +69,9 @@ def __init__(
     def from_array(
         cls,
         array: np.ndarray,
-        uri: Union[str, Path],
-        mode: str = None,
-        format: str = None,
+        uri: Optional[Union[str, Path]] = None,
```
changhiskhan, line 72: now that we have an embedded image version should we separate out the saving to uri portion as a separate method call rather than optional parameters?

python/rikai/types/vision.py
```diff
@@ -184,3 +197,36 @@ def to_numpy(self) -> np.ndarray:
...
+        Returns
+        -------
+        :py:class:`Image` or a list of :py:class:`Image`
+        """
+        if isinstance(box, Box2d):
+            boxes = [box]
+        else:
+            boxes = box
+        crops = []
+        with self.to_pil() as img:
```
changhiskhan, line 226: nit: img -> pil_img for clarity since "image" is overloaded here (Rikai image, PIL image, conceptual image, etc)

python/rikai/types/vision.py
```diff
@@ -184,3 +197,36 @@ def to_numpy(self) -> np.ndarray:
...
+            The bounding box(es) to crop out of this image.
+        format : str, optional
+            The image format to save as
+
+        Returns
+        -------
+        :py:class:`Image` or a list of :py:class:`Image`
+        """
+        if isinstance(box, Box2d):
+            boxes = [box]
```
changhiskhan, line 222: nit: maybe just clearer to have 
```
with self.to_pil().crop(bbox) as patch:
    return Image.from_pil(patch)
```
instead of converting to list then having an extra check later?

changhiskhan (approved):
2 nits

python/rikai/types/vision.py
eddyxu: like `Image.as_external(uri)` or sth?

python/rikai/types/vision.py
eddyxu: sure

python/rikai/types/vision.py
eddyxu: confused. does this suggest `Image.from_pil` to take in a list of patch or sth?

or just handle list and single item differently.

python/rikai/types/vision.py
eddyxu: we can do in a follow up JIRA?

eddyxu: #249

PR #244, opened 2021-04-17
Deduper

smellslikeml:
Implements a Transformer for image deduplication/clustering.

Transforms DataFrame with column of image uris to add column of cluster ids based on hierarchical clustering of grouped images based on SSIM.

`groupIdCol` is used to reduce similarity computations to smaller sets of images likely to be similar. Pinterest used ids from [LSH hashed embeddings](https://medium.com/pinterest-engineering/detecting-image-similarity-using-spark-lsh-and-tensorflow-618636afc939) but the ids can come from hashed metadata such as movie IMDb id.


python/rikai/spark/functions/vision.py
```diff
@@ -280,3 +281,22 @@ def spectrogram_image(
         format=image_format,
         **image_kwargs,
     )
+
+
+@udf(returnType=ArrayType(IntegerType()))
```
changhiskhan, line 286: let's move this to rikai.contrib.spark.vision maybe? how do you feel about putting functionality that's useful but not quite so common to most rikai users to the contrib section @eddyxu ?

python/rikai/spark/functions/vision.py
```diff
@@ -280,3 +281,22 @@ def spectrogram_image(
...
+def img_cluster(img_list, threshold=0.5, img_shape=(300, 300)):
```
changhiskhan, line 287: @smellslikeml does keyword arguments work? like you can't do `img_cluster("image_col", img_shape=(400, 300))` can you?

changhiskhan, line 287: some documentation would be nice, especially describing the return value

python/rikai/spark/functions/vision.py
```diff
@@ -280,3 +281,22 @@ def spectrogram_image(
...
+    from sklearn.cluster import AgglomerativeClustering
+    from skimage.metrics import structural_similarity as ssim
+
+    if len(img_list) == 1:
+        return [0]
+    clustering_model = AgglomerativeClustering(
+        n_clusters=None, distance_threshold=threshold, linkage="ward"
+    )
+    frames = [np.resize(img.to_numpy(), img_shape) for img in img_list]
```
changhiskhan, line 296: from documentation on numpy.resize: "This functionality is therefore not suitable to resize images, or data where each axis represents a separate and distinct entity."

changhiskhan: why call this "frames" as opposed to just images?

python/rikai/spark/functions/vision.py
```diff
@@ -280,3 +281,22 @@ def spectrogram_image(
...
+    clustering_model.fit(
+        np.array(
+            [[ssim(f, ff, multichannel=True) for ff in frames] for f in frames]
```
changhiskhan, line 299: assuming ssim is commutative and self-ssim is a constant, you can save 1/2 of the computations here no?

python/rikai/spark/functions/vision.py
```diff
@@ -280,3 +281,22 @@ def spectrogram_image(
...
+        )
+    )
+    return clustering_model.labels_.tolist()
```
changhiskhan, line 302: some unit tests?

changhiskhan (commented):
looks cool. some comments/questions. thanks for the PR!

python/rikai/spark/functions/vision.py
smellslikeml: Just need the dimensions of reference and test images to be the same for SSIM.

python/rikai/spark/functions/vision.py
smellslikeml: will update using [skimage.transform.resize](https://scikit-image.org/docs/dev/api/skimage.transform.html#skimage.transform.resize)

python/rikai/spark/functions/vision.py
smellslikeml: will do

changhiskhan:
I don't think we can use this as is. I'm going to close this PR and we can create a new one for image clustering as needed

PR #255, opened 2021-04-20
add utility for initializing spark session and initializing rikai sql

changhiskhan:


python/rikai/spark/utils.py
```diff
@@ -43,3 +44,34 @@ def get_default_jar_version(use_snapshot=True):
...
+        )
+        .config(
+            "spark.executor.extraJavaOptions",
+            "-Dio.netty.tryReflectionSetAccessible=true",
+        )
+    )
+    conf = conf or {}
+    for k, v in conf.items():
+        builder = builder.config(k, v)
+    session = builder.master("local[2]").getOrCreate()
```
eddyxu, line 75: is this user facing api? should we make `master` configuable?

eddyxu (approved):
small nits

python/rikai/spark/utils.py
changhiskhan: was just thinking about testing convenience and demo convenience. That being said, it could be a user facing API? what do you think? Would probably have to pull it up to a higher level module like `rikai.spark.init()` or something?

changhiskhan: Going to merge this for now. Can pull it up separately and parameterize master

PR #257, opened 2021-04-20
support min score and do not chop / resize for inference

eddyxu:


python/rikai/contrib/torch/transforms/fasterrcnn_resnet50_fpn.py
```diff
@@ -20,19 +20,21 @@
...
 
+DEFAULT_MIN_SCORE = 0.5
+
 
 def pre_processing(options: Dict[str, Any]) -> Callable:
     return T.Compose(
         [
             uri_to_pil,
-            T.Resize(256),
-            T.CenterCrop(224),
```
eddyxu, line 29: this is useful for training we dont need it for inference.

changhiskhan (approved):

PR #263, opened 2021-04-23
Support pandas < 0.25 in rikai.conf

chunyang:
`pandas.core.config` objects were moved to `pandas._config.config` in Pandas 0.25. Add a try-import to support earlier versions of Pandas.

### DIFF
--- a/python/rikai/conf.py
+++ b/python/rikai/conf.py
@@ -22,6 +22,7 @@ import os
 import tempfile
 
 import pandas as pd
+
 try:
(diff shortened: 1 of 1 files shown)


eddyxu (approved):
The code looks fine. it'd be nice to be able to test it in CI tho

chunyang:
Agree, would be good to test. Let me think about it. The question is how to choose which versions to test. Pandas 0.23 is important for us but not sure for others.

PR #264, opened 2021-04-25
Add udf to retrieve video metadata using ffprobe

changhiskhan:
@smellslikeml had a good prototype for getting the frame rate out using
ffmpeg's ffprobe functionality. This commit generalizes that and makes
it more robust in a distributed context (e.g., error handling).
In addition to the frame rate we're also returning a number of other
metadata attributes like width, height, duration, size, etc.

### DIFF
--- a/python/rikai/spark/functions/vision.py
+++ b/python/rikai/spark/functions/vision.py
@@ -18,12 +18,19 @@
 # Standard library
 import os
 from pathlib import Path
-from typing import Union
+from typing import Optional, Union
 
 # Third Party
 import numpy as np
 from pyspark.sql.functions import udf
-from pyspark.sql.types import ArrayType
+from pyspark.sql.types import (
+    ArrayType,
+    FloatType,
+    IntegerType,
+    StringType,
+    StructField,
+    StructType,
+)
 
 # Rikai
 from rikai.io import copy as _copy
@@ -44,6 +51,8 @@ __all__ = [
     "numpy_to_image",
     "video_to_images",
     "spectrogram_image",
+    "video_metadata",
+    "video_metadata_schema",
 ]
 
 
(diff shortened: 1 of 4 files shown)


smellslikeml (approved):
LGTM

python/rikai/spark/functions/vision.py
```diff
@@ -206,6 +215,152 @@ def video_to_images(
     ]
 
 
+video_metadata_schema = StructType(
+    [
+        StructField(
+            "error",
+            StructType(
```
eddyxu, line 222: so, do we want to advocate this error handling pattern? the one that returns either data or error?

Or , should we just return `data` part , while put `error` into logs? like if anything was wrong, return `null` instead?  It makes it easier to be integrate with other part of the systems, i.e, nested UDFs.

python/rikai/spark/functions/vision.py
```diff
@@ -206,6 +215,152 @@ def video_to_images(
...
+    --------
+    The following returns the fps rounded to the nearest integer
+    ```
+    import rikai.spark.functions as RF
+    (spark.createDataFrame([(VideoStream(<uri>),)], ['video'])
+          .withColumn('meta', RF.video_metadata('video'))
+          .select('meta.data.frame_rate'))
+    ```
+    """
+    probe_result = _probe(video)
```
eddyxu, line 281: This will raise exception, which will not be captured in the return?

python/rikai/spark/functions/vision.py
eddyxu: is this a commonly used pattern in spark?

changhiskhan: It's a Scala pattern (try).

Putting the errors in the logs makes it harder to diagnose the error for a big dataset. e.g., i might want to see 1) how many links still point to something 2) how many links point to random files 3) how many links point to audio only tracks, 4) how many files point to video files that are mal-formed/corrupted. 

For ease of use, an expr like "output_col.data.width" will return null for the rows where there are errors and no data so you don't need complicated handling to get a particular attribute if you just want the data.

python/rikai/spark/functions/vision.py
changhiskhan: This will raise an exception if ffmpeg is not installed. In this case we just want to short circuit. The idea with the error handling is that each individual ffprobe call is to a commandline tool where the errors/exceptions often looks weird/opaque to python/scala.

python/rikai/spark/functions/vision.py
eddyxu: i still dont feel strong with this.  it is like from the relational algebra's aspect, that a table does not only contains data, it also contains "what happened during the time when the data is generated".

in `Apache Beam`, we can do something like

```
data, error = pipeline | Map(func)
```
i dont know what is the good equivalent here.


Can we at least make the schema like:

```
StructType([
        StructField("width", IntegerType(), True),
        StructField("height", IntegerType(), True),
        StructField("num_frames", IntegerType(), True),
        StructField("duration", FloatType(), True),
        StructField("bit_rate", IntegerType(), True),
        StructField("frame_rate", IntegerType(), True),
        StructField("codec", StringType(), True),
        StructField("size", IntegerType(), True),
        StructField("_error", StructType([
                StructField("message", StringType(), True),
                StructField("stderr", StringType(), True),
                StructField("probe", StringType(), True),
            ]
           , True),
        ),
    ]
),
```

In BigQuery, there are a few system added / hidden column named with `_` as prefix.

python/rikai/spark/functions/vision.py
```diff
@@ -206,6 +215,152 @@ def video_to_images(
...
+        )
+    except Exception as e:
+        return _error(str(e))
+
+
+def _probe(video):
+    try:
+        import ffmpeg
+    except ImportError:
+        raise ValueError(
```
eddyxu, line 319: why not re-raise `ImportError` like

```raise ImportError(...) from e```?

python/rikai/spark/functions/vision.py
changhiskhan: yeah i can do that. I much prefer having clean separation between data and errors, but I recognize that this pattern is only familiar in functional programming and not really for ML/data people as much, esp when they're interacting with DataFrame APIs.

python/rikai/spark/functions/vision.py
changhiskhan: copy-paste

eddyxu (approved):
LGTM

PR #268, opened 2021-04-26
Support pin memory in torch runner DataLoader

### LINKED ISSUE #267: add pin_memory option
not a priority but could offer some additional performance improvement when using the gpu

eddyxu:
Closes #267

### DIFF
--- a/python/rikai/spark/sql/codegen/pytorch.py
+++ b/python/rikai/spark/sql/codegen/pytorch.py
@@ -25,6 +25,7 @@ from rikai.torch.pandas import PandasDataset
 
 DEFAULT_NUM_WORKERS = 8
 DEFAULT_BATCH_SIZE = 4
+DEFAULT_PIN_MEMORY = True
 
 
 def generate_udf(spec: "rikai.spark.sql.codegen.base.ModelSpec"):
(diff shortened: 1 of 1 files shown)


python/rikai/spark/sql/codegen/pytorch.py
```diff
@@ -63,6 +65,7 @@ def torch_inference_udf(
                     dataset,
                     batch_size=batch_size,
                     num_workers=num_workers,
+                    pin_memory=pin_memory,
                 ):
                     batch = batch.to(device)
```
changhiskhan, line 70: I *think* if pin_memory is set to be true, you also need to do `batch.to(device, non_blocking=True)` for performance gains.

python/rikai/spark/sql/codegen/pytorch.py
```diff
@@ -63,6 +65,7 @@ def torch_inference_udf(
                     dataset,
                     batch_size=batch_size,
                     num_workers=num_workers,
+                    pin_memory=pin_memory,
```
changhiskhan, line 68: also don't you have to actually call `Tensor.pin_memory` somewhere?

python/rikai/spark/sql/codegen/pytorch.py
```diff
@@ -25,6 +25,7 @@
 
 DEFAULT_NUM_WORKERS = 8
 DEFAULT_BATCH_SIZE = 4
+DEFAULT_PIN_MEMORY = True
```
changhiskhan, line 28: pytorch has pin memory defaulting to False because in some instances it can result in swap/freeze that's hard to diagnose

PR #272, opened 2021-04-26
Scene segmentation

changhiskhan:
Add contrib module to use pyscenedetect to split scenes.


kaldarabsah (approved):

python/rikai/contrib/s3.py
```diff
@@ -0,0 +1,49 @@
...
+OpenCV cannot open s3 url's directly. Instead we make a presigned s3 url.
+"""
+import logging
+from urllib.parse import urlparse
+
+try:
+    import boto3
+    from botocore.exceptions import ClientError
+except ImportError as e:
+    raise ImportError("Please `pip install awscli` to use s3 utils") from e
```
eddyxu, line 25: should we put `boto3` into `rikai[aws]` package? and  here , why did we ask to install `awscli` while it is `boto3` missing?

python/rikai/contrib/s3.py
```diff
@@ -0,0 +1,49 @@
+#  Copyright (c) 2021 Rikai Authors
```
eddyxu, line 1: this could be put into main rikai repo instead of `contrib` tho.

python/rikai/contrib/video/functions.py
```diff
@@ -0,0 +1,150 @@
...
+        if raise_on_error:
+            raise e
+        else:
+            return None
+    finally:
+        if video_manager is not None:
+            video_manager.release()
+
+
+def normalize_uri(uri: Union[str, list]):
```
eddyxu, line 141: should this merge with `rikai.io`?

smellslikeml (commented):
Exposing the `downscale_factor` parameter may improve performance speed.

python/rikai/contrib/video/functions.py
changhiskhan: I put it here because it's not clear how we should merge it into rikai.io, For opencv we have to generate a pre-signed URL, but for PIL maybe we want to copy the file first OR open a file handle that's wrapped around boto3 API call.

python/rikai/contrib/s3.py
changhiskhan: yeah, was just putting it here til we have a "rikai blessed" s3 interface for all the things.

python/rikai/contrib/s3.py
changhiskhan: i remember you removed that aws extras package for a specific purpose? If we can add it back that makes it easier

eddyxu: yea, i removed it because i found that we can use `pyarrow` for s3 access.  If you need `boto3` here, lets just add them

eddyxu (approved):
lgtm

changhiskhan:
> Exposing the `downscale_factor` parameter may improve performance speed.

if the argument is none it'll already do some automatic downscaling. I discussed with khaldun if we find that we need to have more fine-grained control, we can just make a udf generator that can deal with keyword arguments easily rather than putting that in the udf signature itself (udf's don't work with keyword arguments)

PR #273, opened 2021-04-27
Support crop as a pyspark UDF

eddyxu:


python/rikai/spark/functions/vision.py
```diff
@@ -425,3 +431,28 @@ def spectrogram_image(
...
+    list of :py:class:`Image`
+
+    Examples
+    --------
+
+    >>> spark.sql(\"\"\"SELECT crop(image, boxes) as patches FROM image\"\"\")
+
+    """
+
+    return img.crop(boxes)
```
chunyang, line 458: Might I suggest that we name the Rikai operation something other than `crop`? `crop` to me implies that you crop a single input image to to get a single output image.

How about something like `extract_patches`? Though this might be confused with [TensorFlow's `extract_patches`](https://www.tensorflow.org/api_docs/python/tf/image/extract_patches), which is semantically different (it extracts regularly spaced patches rather than arbitrary bounding boxes).

python/rikai/spark/functions/vision.py
```diff
@@ -425,3 +431,28 @@ def spectrogram_image(
...
+@udf(returnType=ArrayType(ImageType()))
+def crop(img: Image, boxes: List[Box2d]):
+    """Crop images from the original image
+
+    Parameters
+    ----------
+    img : Image
+        The base image to crop from.
+    boxes : List[Box2d]
+        A list of bounding boxes to specify the area to crop.
```
chunyang, line 445: Can you add documentation on whether these bounding boxes are normalized? i.e., should they have range [0, 1] or range [0, image_width]?

chunyang (commented):
Some opinions :)

python/rikai/spark/functions/vision.py
eddyxu: Good point.

python/rikai/spark/functions/vision.py
eddyxu: ok, let me take a look of this api

eddyxu: This is more aligned with https://www.tensorflow.org/api_docs/python/tf/image/crop_to_bounding_box ?

chunyang: Sort of, `crop_to_bounding_box` produces only one patch though.

eddyxu: `extract_patch` is making patches with the same size and the same strides. It is not cut patch out of an image with specified areas. 

i can rename this as `make_patches(image, boxes, padding=[1, 2, 3, 4])`?

python/rikai/spark/functions/vision.py
eddyxu: Maybe we should just use normalized box across rikai. what do you think @changhiskhan

PR #278, opened 2021-05-06
Implement CREATE MODEL IF NOT EXISTS

da-tubi:


src/main/antlr4/org/apache/spark/sql/ml/parser/RikaiExtSqlBase.g4
```diff
@@ -21,6 +21,7 @@ singleStatement
 
 statement
     : CREATE (OR REPLACE)? MODEL model=qualifiedName
+      (IF NOT EXISTS)?
```
eddyxu, line 24: should we make `(IF NOT EXISTS)` before `MODEL`?

`IF NOT EXISTS` and `OR REPLACE` can not be chosen at the same time.

i.e., https://cloud.google.com/bigquery-ml/docs/reference/standard-sql/bigqueryml-syntax-create

PR #279, opened 2021-05-12
pin antlr4Version

da-tubi:
This pr does not solve existing bugs, it prevents future potential bugs.

If the ANTRL version does not match, the following warnings are printed:

> ANTLR Tool version 4.7 used for code generation does not match the current runtime version 4.8ANTLR Tool version 4.7 used for code generation does not match the current runtime version 4.8

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -91,6 +91,9 @@ fork in Test := true
 
 antlr4PackageName in Antlr4 := Some("ai.eto.rikai.sql.spark.parser")
 antlr4GenVisitor in Antlr4 := true
+// explicitly pin antlr4Version to avoid potential bugs
+// antlr4Version should be the same with the one bundled with Apache Spark
+antlr4Version in Antlr4 := "4.8-1"
 
 enablePlugins(Antlr4Plugin)
 


build.sbt
```diff
@@ -91,6 +91,9 @@ fork in Test := true
 
 antlr4PackageName in Antlr4 := Some("ai.eto.rikai.sql.spark.parser")
 antlr4GenVisitor in Antlr4 := true
+// explicitly pin antlr4Version to avoid potential bugs
+// antlr4Version should be the same with the one bundled with Apache Spark
+antlr4Version in Antlr4 := "4.8-1"
```
eddyxu, line 96: this is what used in `spark 3.0+`?
For reference, ubuntu 20.04 seems to use 4.7.x while `homebrew` on OSX installs `4.9.x`.

Btw, should we also pin the version here: https://github.com/eto-ai/rikai/blob/master/python/setup.py#L50

i think this was where the error message came from.

da-tubi: Yes. Apache Spark 3.1.1 is using Antlr 4.8-1. And `sbt-antlr` 0.8.3 happens to use Antlr 4.8-1.

If you manually pin antlr4Version to 4.7. The warning message in the PR desc will be displayed.

da-tubi: "the error message" is manually produced by me to show why we need to use the consistent antlr version.

da-tubi: For python, we'd better make it consistent. The generated python code are from Antlr 4.7. But the antlr4-python3-runtime is 4.9.x. How about complete it in another PR?

eddyxu (approved):

eddyxu:
Please make sure merge this around the same time with the python version bump. 

IMHO, it'd better to just bump antlr version in python in this PR.  Up to you.

eddyxu:
Duplicated to #299

PR #283, opened 2021-05-15
Support http auth to open_uri

### LINKED ISSUE #280: Feature request: Support passing HTTP request headers to open_uri
Sometimes headers are required to download assets via HTTP. An example of this is where the endpoint requires authentication via the Authorization header. We should allow [`rikai.io.open_uri()`](https://github.com/eto-ai/rikai/blob/c69f8169882879026f8d91fac7d885aed09c968e/python/rikai/io.py#L127) to access these assets.

eddyxu:
Closes #280


chunyang (approved):
LGTM thanks!

PR #284, opened 2021-05-16
Support returning primitives in ML_PREDICT in torch flavor 

eddyxu:


changhiskhan (commented):
the code looks good. The main question I have is whether we want to provide model packages that do these types of functionality OR provide aggregates on top of the predict outcome.

e.g., if we want the count and mode, do we have to run two separate models where everything is the same except for post-processing?

eddyxu:
This was not for providing "object_counts“ as the objective of this PR.  This PR is allow pytorch runner able to run a model that yield primitive types. 

Use `object_counts` for a testing example

changhiskhan (approved):

PR #286, opened 2021-05-31
Vectorize calculating iou with a batch of Box2d

eddyxu:


python/rikai/types/geometry.py
```diff
@@ -263,22 +263,40 @@ def area(self) -> float:
...
         return self.width * self.height
 
+    @staticmethod
+    def _area(arr: np.ndarray) -> np.array:
+        """Calculate box area from an array of boxes."""
+        return np.maximum(0, arr[:, 2] - arr[:, 0]) * np.maximum(
+            0, arr[:, 3] - arr[:, 1]
+        )
+
     def iou(self, other: "Box2d") -> float:
```
changhiskhan, line 273: update type hints?

python/rikai/types/geometry.py
```diff
@@ -263,22 +263,40 @@ def area(self) -> float:
...
         """Compute intersection over union(IOU)."""
         assert isinstance(
-            other, Box2d
+            other, (Box2d, list)
```
changhiskhan, line 276: what about like a tuple or array of Box2d?

changhiskhan (approved):
2 nits

python/rikai/types/geometry.py
eddyxu: done

PR #287, opened 2021-06-03
Allow taking empty array of bounding box to run iou

eddyxu:


python/rikai/types/geometry.py
```diff
@@ -281,6 +281,8 @@ def iou(
             other_arr = np.array([other])
         else:
             other_arr = np.array(other)
+        if other_arr.size == 0:
+            return np.zeros(0)
```
changhiskhan, line 285: what's the difference between np.zeros(0) and np.empty?

changhiskhan (approved):

python/rikai/types/geometry.py
chunyang: `np.empty` doesn't initialize the array elements to anything. Since in this case there are no array elements it doesn't matter, but I think `np.empty(0)` is cleaner.

chunyang (approved):

python/rikai/types/geometry.py
eddyxu: :this:

PR #292, opened 2021-06-25
Error Handling for MLflow protocol with 2 forward slashes

da-tubi:
## Why `mlflow://` does not work
``` python
>>> from urllib.parse import urlparse
>>> p1 = urlparse("mlflow:/a/1")
>>> p2 = urlparse("mlflow://b/2")
>>> p3 = urlparse("mlflow:///c/3")
>>> p4 = urlparse("mlflow:////d/4")
>>> p1
ParseResult(scheme='mlflow', netloc='', path='/a/1', params='', query='', fragment='')
>>> p2
ParseResult(scheme='mlflow', netloc='b', path='/2', params='', query='', fragment='')
>>> p3
ParseResult(scheme='mlflow', netloc='', path='/c/3', params='', query='', fragment='')
>>> p4
ParseResult(scheme='mlflow', netloc='', path='//d/4', params='', query='', fragment='')
```

## How to test locally
```
sbt publishLocal
# Install python package
cd python
pip install -e . # pip install -e .[all] to install all optional extras (see "Install from pypi")
pytest tests/spark/sql/codegen/test_mlflow_registry.py -k test_mlflow_model_from_model_version
```

### DIFF
--- a/python/rikai/spark/sql/codegen/mlflow_registry.py
+++ b/python/rikai/spark/sql/codegen/mlflow_registry.py
@@ -217,7 +217,8 @@ class MlflowRegistry(Registry):
         parsed = urlparse(uri)
         if parsed.netloc:
             raise ValueError(
-                "URI with 2 forward slashes is not supported, try URI with 1 slash instead"
+                "URI with 2 forward slashes is not supported, "
+                "try URI with 1 slash instead"
             )
         if not parsed.scheme:
             raise ValueError("Scheme must be mlflow. How did you get here?")
--- a/python/rikai/spark/sql/codegen/mlflow_registry.py
+++ b/python/rikai/spark/sql/codegen/mlflow_registry.py
@@ -215,6 +215,8 @@ class MlflowRegistry(Registry):
         uri = raw_spec.getUri()
         logger.info(f"Resolving model {name} from {uri}")
         parsed = urlparse(uri)
+        if parsed.netloc:
+            raise ValueError("MLflow protocol with 2 forward slashes is not supported")
         if not parsed.scheme:
             raise ValueError("Scheme must be mlflow. How did you get here?")
         parts = parsed.path.strip("/").split("/", 1)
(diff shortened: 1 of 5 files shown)


da-tubi:
CI is blocked by other tests, this PR looks good to me now.

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -215,6 +215,8 @@ def resolve(self, raw_spec):
         uri = raw_spec.getUri()
         logger.info(f"Resolving model {name} from {uri}")
         parsed = urlparse(uri)
+        if parsed.netloc:
```
eddyxu, line 218: Could you add some document here, i.e. including the model url spec in mlflow?

Wanted to understand that `mlflow` does not actually support netloc field?

darcy-shen: The reason why `mlflow://` does not work is that our implementation does not allow `mlflow://`

I think it could be regarded as bug. And let fix it.

changhiskhan:
@da-tubi this is intended behavior. the specification is `<schema>://<host>/path/to/resource` so if you have a URL like p2, `b` should be parsed as the host according to the uri standard no?

changhiskhan:
we should stick to the uri standard specification. If I'm misreading it please let me know and we can fix it.

darcy-shen:
@changhiskhan  It was reverted now.

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -215,6 +215,8 @@ def resolve(self, raw_spec):
...
+            raise ValueError("URI with 2 forward slashes is not supported")
```
changhiskhan, line 219: The error message should tell the user how to do the right thing. e.g., you might want to add "Try <user input but with 1 slash> or <user input but with 3 slashes> instead" to the message

changhiskhan (changes requested):

python/rikai/spark/sql/codegen/mlflow_registry.py
da-tubi: Improved.

PR #302, opened 2021-08-03
Crop image udf

eddyxu:


python/rikai/spark/functions/vision.py
```diff
@@ -425,3 +427,24 @@ def spectrogram_image(
...
+
+@udf(returnType=ArrayType(ImageType(), True))
+def crop(img: Image, box: Union[Box2d, List[Box2d]]):
+    """Crop image specified by the bounding boxes, and returns the cropped
+    images.
+
+    Parameters
+    ----------
+    img : Image
+        An image object.
```
changhiskhan, line 440: Nit: "The image to be cropped"

python/rikai/spark/functions/vision.py
```diff
@@ -425,3 +427,24 @@ def spectrogram_image(
...
+    box : Box2d or List of Box2d
+        One bound-2d box or a list of such boxes
```
changhiskhan, line 442: Nit: maybe add that if a list of bounding boxes are given as input, the output will also be a list/array?

python/rikai/types/vision.py
```diff
@@ -160,7 +160,11 @@ def display(self, **kwargs):
             return Image(fobj.read(), **kwargs)
 
     def __repr__(self) -> str:
-        return f"Image(uri={self.uri})"
+        return (
```
changhiskhan, line 163: Nit: it's probably more readable as just a regular if/else instead of doing it inline

python/rikai/io.py
```diff
@@ -164,6 +165,12 @@ def open_uri(
         # This is a local file
         return open(uri, mode=mode)
     elif parsed_uri.scheme in ("http", "https"):
+        if http_headers is None:
+            http_headers = {}
+        if "User-Agent" not in http_headers:
```
changhiskhan, line 170: Just curious why this is needed (or what other values we should expect)?

changhiskhan (approved):
LGTM

Sorry for the delay

python/rikai/io.py
eddyxu: Now wikipedia expects a http request has agent header.  Useful in the tests for now.

I leave this to customizable so that some of the downstream customers can customize it for access control?

python/rikai/spark/functions/vision.py
eddyxu: It will return a list of images. Because udf can only support one type of output schema.

PR #309, opened 2021-08-28
Provide a gcs_tmpdir fixture to enable GCS based tests

eddyxu:
Similar to what `s3_tmpdir` does, this offers the capability to run tests on gcs.


python/rikai/io.py
```diff
@@ -25,9 +25,10 @@
 import requests
 from pyarrow import fs
 
+import rikai.conf
```
eddyxu, line 28: It was done via `make fix`

python/tests/conftest.py
```diff
@@ -58,3 +73,41 @@ def resnet_model_uri(tmp_path_factory):
...
+def gcs_tmpdir() -> str:
+    """Create a temporary Google Cloud Storage (GCS) directory to test dataset.
+
+    To enable GCS test, it requires both the AWS credentials,
+    as well as `RIKAI_TEST_GCS_BUCKET` being set.
+    """
+
+    bucket = os.environ.get("RIKAI_TEST_GCS_BUCKET", None)
+    if bucket is None:
+        pytest.skip("skip test. RIKAI_TEST_GCS_BUCKET is not set")
```
chunyang, line 103: nit
```suggestion
        pytest.skip("skipping test. RIKAI_TEST_GCS_BUCKET is not set")
```

python/tests/conftest.py
```diff
@@ -58,3 +74,65 @@ def resnet_model_uri(tmp_path_factory):
     model_uri = tmp_path / "resnet.pth"
     torch.save(resnet, model_uri)
     return model_uri
+
+
+@pytest.fixture
+def gcs_tmpdir() -> str:
+    """Create a temporary Google Cloud Storage (GCS) directory to test dataset.
+
+    To enable GCS test, it requires both the AWS credentials,
```
chunyang, line 83:
```suggestion
    To enable GCS test, it requires both the GCP credentials,
```

python/tests/conftest.py
```diff
@@ -26,6 +32,8 @@
 from rikai.spark.sql import init
 from rikai.spark.utils import init_spark_session
 
+random.seed(time.time())
```
chunyang, line 35: Isn't the default seed already random?

python/tests/conftest.py
```diff
@@ -58,3 +74,65 @@ def resnet_model_uri(tmp_path_factory):
...
+    To enable GCS test, it requires both the AWS credentials,
+    as well as `RIKAI_TEST_GCS_BUCKET` being set.
+
+    Examples
+    --------
+
+    .. code-block:: bash
+
+        $ export GOOGLE_APPLICATION_CREDENTIALS=/path/to/key.json
+        $ export RIKAI_TEST_GCS_BUCKET=gs://bucket
```
chunyang, line 92: Should bucket also contain the `gs://` part? I got confused reading line 125 where I didn't realize the `bucket` variable was prefixed by `gs://`.

python/tests/conftest.py
eddyxu: Ah, you are right, Thanks @chunyang !

python/tests/conftest.py
eddyxu: Good point, would the env var to be named as `RIKAI_TEST_GCS_PATH` be more appropriated?

eddyxu: personally i'd prefer `gs://...`, it is more explicit , and this is the form that used in spark and python reader.

python/tests/conftest.py
eddyxu: Done

python/tests/conftest.py
eddyxu: Done

python/tests/conftest.py
chunyang: Oh I meant I don't think you have to call `random.seed()` at all lol

eddyxu:
Have run

```
RIKAI_TEST_GCS_BUCKET=gs://bucket  pytest python/tests/parquet/test_dataset.py -s -k gcs
```

python/tests/conftest.py
chunyang: Okay, can you add a check that the yielded `temp_dir` contains a `gs://` prefix then? Otherwise it's possible for someone to specify `export RIKAI_TEST_GCS_BUCKET=bucket` and end up writing to a local file `bucket/xyz123` when doing `df.write.format("rikai").save(tmpdir)`.

eddyxu: I will add a check on the  input then. So it enforces user to use `gs://`

python/rikai/spark/utils.py
```diff
@@ -54,7 +54,20 @@ def init_spark_session(conf=None, app_name="rikai", rikai_version=None):
...
         .config(
-            "spark.jars.packages", "ai.eto:rikai_2.12:{}".format(rikai_version)
+            "spark.jars.packages",
+            ",".join(
+                [
+                    "ai.eto:rikai_2.12:{}".format(rikai_version),
+                ]
+            ),
+        )
+        .config(
```
changhiskhan, line 64: do we want to have it so that these configurations are only applied for testing GCP related functionality? Could potentially cause problems if a user is on AWS and they're using these non-standard jars?

python/tests/conftest.py
```diff
@@ -58,3 +72,66 @@ def resnet_model_uri(tmp_path_factory):
...
+
+    References
+    ----------
+    https://cloud.google.com/dataproc/docs/concepts/connectors/cloud-storage
+    https://cloud.google.com/dataproc/docs/concepts/iam/iam
+    """
+
+    base_url = os.environ.get("RIKAI_TEST_GCS_URL", None)
+    if base_url is None:
+        pytest.skip("Skipping test. RIKAI_TEST_GCS_URL is not set")
```
changhiskhan, line 101: how would we distinguish between an actual error vs skipping here? e.g., if the env var is misconfigured?

python/rikai/spark/utils.py
eddyxu: I dont think that the production deployment can use this function at all. For example, jars, including rikai , should be located with the other jars in the OS, deployed via `spark-defaults.conf`?

For example, the above `spark.jar.packages` might overwrite deployed values as well. 

Should we mark this as internal use only?

python/tests/conftest.py
eddyxu: `pytest.skip` does not break the tests, while `raise ValueError` does.

The potential mistakely skipped case is that we offer a gs://url but does not have permission to read it. We can probably check the error message from the exception?

If we want to put hard enforcement on it, we can add an extra marker, like `--gcs-test`, and check it here.

```
if not pytest.markers.get("gcs_test"):
    return

assert <any error later>
```

changhiskhan (approved):

PR #315, opened 2021-09-13
Support ROS pointcloud type

eddyxu:


eddyxu:
A question to discuss is, whether we should put this as a `rikai-contiribute` or `rikai-ros` package. @changhiskhan

Issue #319, opened 2021-09-22
pre/post processors are overrided by pre/post processors recorded in MLflow tags


darcy-shen:
https://github.com/eto-ai/rikai/blob/00f3ba737f5ca94f57e678a20fd622944b622625/python/rikai/spark/sql/codegen/mlflow_registry.py#L246-L258

It means, when we set the processors using MLflow tags, we can not customize it using:

```
create model xxx_model
preprocessor 'x.y.z.pre_process'
using 'mlflow:/xxxx'
```

PR #320, opened 2021-09-22
pre/post processors: specs should take precedence over tags

da-tubi:
Fix https://github.com/eto-ai/rikai/issues/319


changhiskhan (approved):
Makes sense thanks

PR #325, opened 2021-09-26
Add mixin to convert Rikai types to dict

changhiskhan:
convenience mixin to help convert Rikai types to dict. This makes it easier for serde/display.

### DIFF
--- a/python/rikai/mixin.py
+++ b/python/rikai/mixin.py
@@ -16,7 +16,7 @@
 """
 from __future__ import annotations
 
-from abc import ABC, ABCMeta, abstractmethod
+from abc import ABC, abstractmethod
 from io import BytesIO
 from pathlib import Path
 from typing import BinaryIO, Optional, Union
@@ -47,7 +47,7 @@ class ToPIL(ABC):
         pass
 
 
-class ToDict(metaclass=ABCMeta):
+class ToDict(ABC):
     """ToDict Mixin"""
 
     @abstractmethod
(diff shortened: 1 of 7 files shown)


python/rikai/mixin.py
```diff
@@ -47,6 +47,14 @@ def to_pil(self) -> "PIL.Image.Image":
         pass
 
 
+class ToDict(metaclass=ABCMeta):
```
eddyxu, line 50: Why use ABCMeta instead of just ABC here like the rest of the codebase?

python/rikai/types/vision.py
```diff
@@ -200,6 +200,11 @@ def to_numpy(self) -> np.ndarray:
         with self.to_pil() as pil_img:
             return np.asarray(pil_img)
 
+    def to_dict(self) -> dict:
+        if self.is_embedded:
```
eddyxu, line 204: No strong opinion here. Just general question, why not return `{"data": self.data, "uri": null}` here?
Btw, should we do a base64 encoding on the data?

python/rikai/mixin.py
changhiskhan: muscle memory

python/rikai/types/vision.py
changhiskhan: yeah, added base64 encoding.
return schema: i don't have a strong preference either just wanted to avoid sending extra keys that we know are null given the context that it's table data so presumably the data schema is known/declared explicitly elsewhere

eddyxu: Ok, that makes sense.

eddyxu (approved):
LGTM thanks!

PR #326, opened 2021-09-27
Support sklearn flavor under the mlflow registery

da-tubi:
Here are sample usages:

https://github.com/da-tubi/rikai-example/pull/1

### DIFF
--- a/python/rikai/spark/sql/codegen/sklearn.py
+++ b/python/rikai/spark/sql/codegen/sklearn.py
@@ -13,11 +13,11 @@
 #  limitations under the License.
 
 from typing import Iterator
-import pandas as pd
-import numpy as np
 
-from pyspark.sql.types import StructType
+import numpy as np
+import pandas as pd
 from pyspark.sql.functions import pandas_udf, PandasUDFType
+from pyspark.sql.types import StructType
 
 
 def generate_udf(spec: "rikai.spark.sql.codegen.base.ModelSpec"):
--- a/python/rikai/spark/sql/codegen/sklearn.py
+++ b/python/rikai/spark/sql/codegen/sklearn.py
@@ -32,7 +32,10 @@ def generate_udf(spec: "rikai.spark.sql.codegen.base.ModelSpec"):
     -------
     A Spark Pandas UDF.
     """
-    def sklearn_inference_udf(iter: Iterator[pd.Series]) -> Iterator[pd.Series]:
+
+    def sklearn_inference_udf(
+        iter: Iterator[pd.Series],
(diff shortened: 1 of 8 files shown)


python/rikai/spark/sql/codegen/sklearn.py
```diff
@@ -0,0 +1,46 @@
...
+
+from typing import Iterator
+import pandas as pd
+import numpy as np
+
+from pyspark.sql.types import StructType
+from pyspark.sql.functions import pandas_udf
+
+
+def generate_udf(spec: "rikai.spark.sql.codegen.base.ModelSpec"):
```
changhiskhan, line 23: Can you add a unit test for this? LogisticRegression or RandomForest would probably be the most commonly encountered

python/rikai/spark/sql/codegen/sklearn.py
```diff
@@ -0,0 +1,46 @@
...
+    A Spark Pandas UDF.
+    """
+
+    def sklearn_inference_udf(
+        iter: Iterator[pd.Series],
+    ) -> Iterator[pd.Series]:
+        model = spec.load_model()
+        for series in list(iter):
+            assert series.shape == (1,)
+            X = [x.tolist() for x in series[0]]
```
changhiskhan, line 42: Why do you need to convert to list? Big performance hit no?

python/rikai/spark/sql/codegen/sklearn.py
```diff
@@ -0,0 +1,46 @@
...
+            Y = model.predict(X)
+            yield pd.Series([Y.tolist()])
```
changhiskhan, line 44: This will return a Series with one entry that is a list of all predicted y values. Is that what you want?

python/rikai/spark/sql/codegen/sklearn.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+
+    def sklearn_inference_udf(
+        iter: Iterator[pd.Series],
+    ) -> Iterator[pd.Series]:
+        model = spec.load_model()
+        for series in list(iter):
+            assert series.shape == (1,)
+            X = [x.tolist() for x in series[0]]
+            Y = model.predict(X)
```
changhiskhan, line 43: Are we try to predict multiple dependent variables here? The common convention is that X is capitalize because it’s a matrix but y is not because it’s a column vector.

changhiskhan (changes requested):
Cool feature! Look forward to seeing it completed

changhiskhan (approved):

PR #338, opened 2021-10-18
Predict spark ml

Renkai:
Add spark ml support for Rikai, examples are preparing.

### DIFF
--- a/python/rikai/spark/sql/codegen/spark.py
+++ b/python/rikai/spark/sql/codegen/spark.py
@@ -15,6 +15,9 @@ from typing import Iterator
 
 import pandas as pd
 from pyspark.sql.functions import pandas_udf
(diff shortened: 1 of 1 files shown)


python/rikai/spark/sql/codegen/base.py
```diff
@@ -193,6 +193,10 @@ def udf_from_spec(spec: ModelSpec):
     elif spec.flavor == "sklearn":
         from rikai.spark.sql.codegen.sklearn import generate_udf
 
+        return generate_udf(spec)
+    elif spec.flavor == "sparkml":
```
darcy-shen, line 197: `sparkml` -> `spark`

https://www.mlflow.org/docs/latest/models.html#spark-mllib-spark

It would be better to keep the flavor name as it is in MLflow.

Issue #339, opened 2021-10-20
Vectorization of the IOU computation between two lists of box2ds


eddyxu:
For a function signature

```
def iou(box1: list[box2d],  box2: list[box2d]) -> np.ndarray:
   ...
```

For two lists of box2ds, which have the length of `N`, and `M` respectively, this function should return a `N*M` matrix, each element is the iou value (float,`[0, 1]`).

Renkai:
Code for plan version
```
    @staticmethod
    def iou_matrix2(box_list1: List[Box2d], box_list2: List[Box2d]) -> np.ndarray:
        def getvalue(boxa: Box2d, boxb: Box2d):
            xmin = max(boxa.xmin, boxb.xmin)
            ymin = max(boxa.ymin, boxb.ymin)
            xmax = min(boxa.xmax, boxb.xmax)
            ymax = min(boxa.ymax, boxb.ymax)
            inter_area = max(0.0, xmax - xmin) * max(0.0, ymax - ymin)
            return inter_area / (boxa.area + boxb.area - inter_area)

        result = list()
        for a in box_list1:
            row = list()
            for b in box_list2:
                row.append(getvalue(a, b))
            result.append(np.asarray(row))
        return np.asarray(result)
```

Renkai:
I think this issue has been solved now.

PR #341, opened 2021-10-21
Iou two lists

Renkai:
This pr solves https://github.com/eto-ai/rikai/issues/339
I use the name `iou_matrix` not `iou` because the python interpreter believes with name `iou` the type would be ambiguous with another function https://github.com/eto-ai/rikai/blob/38457fc7657ab46d815a5f1488f8c55aab5e7012/python/rikai/types/geometry.py#L302

### DIFF
--- a/python/benchmark/types/iou_benchmark.py
+++ b/python/benchmark/types/iou_benchmark.py
@@ -22,7 +22,9 @@ import numpy as np
 from rikai.types import Box2d
 
 
-def iou_matrix_naive_version(box_list1: Sequence[Box2d], box_list2: Sequence[Box2d]) -> np.ndarray:
+def iou_matrix_naive_version(
+    box_list1: Sequence[Box2d], box_list2: Sequence[Box2d]
+) -> np.ndarray:
     def getvalue(boxa: Box2d, boxb: Box2d):
         xmin = max(boxa.xmin, boxb.xmin)
         ymin = max(boxa.ymin, boxb.ymin)
@@ -51,13 +53,17 @@ def benchmark(list1_len: int, list2_len: int, times: int):
     list1 = [a_random_box2d() for _ in range(0, list1_len)]
     list2 = [a_random_box2d() for _ in range(0, list2_len)]
 
-    naive_seconds = timeit.timeit(lambda: iou_matrix_naive_version(list1, list2), number=times)
-    vectorized_seconds = timeit.timeit(lambda: Box2d.ious(list1, list2), number=times)
+    naive_seconds = timeit.timeit(
+        lambda: iou_matrix_naive_version(list1, list2), number=times
+    )
+    vectorized_seconds = timeit.timeit(
+        lambda: Box2d.ious(list1, list2), number=times
+    )
 
     print("naive method cost {} seconds".format(naive_seconds))
     print("vectorized method cost {} seconds".format(vectorized_seconds))
     print("naive/vectorized {}".format(naive_seconds / vectorized_seconds))
 
 
-if __name__ == '__main__':
+if __name__ == "__main__":
     benchmark(*[int(x) for x in sys.argv[1:]])
--- a/python/benchmark/types/iou_benchmark.py
+++ b/python/benchmark/types/iou_benchmark.py
@@ -1,0 +1,61 @@
+import datetime
+import random
+import sys
+from typing import Sequence
+
+import numpy as np
+from datetime import datetime
+from rikai.types import Box2d
+
+
+def iou_matrix_naive_version(box_list1: Sequence[Box2d], box_list2: Sequence[Box2d]) -> np.ndarray:
+    def getvalue(boxa: Box2d, boxb: Box2d):
+        xmin = max(boxa.xmin, boxb.xmin)
+        ymin = max(boxa.ymin, boxb.ymin)
+        xmax = min(boxa.xmax, boxb.xmax)
+        ymax = min(boxa.ymax, boxb.ymax)
+        inter_area = max(0.0, xmax - xmin) * max(0.0, ymax - ymin)
+        return inter_area / (boxa.area + boxb.area - inter_area)
+
+    result = list()
+    for a in box_list1:
+        row = list()
+        for b in box_list2:
+            row.append(getvalue(a, b))
+        result.append(np.asarray(row))
+    return np.asarray(result)
+
+
+def benchmark(list1_len: int, list2_len: int, times: int):
+    def a_random_box2d():
+        x_min = random.uniform(0, 1)
+        y_min = random.uniform(0, 1)
+        x_max = random.uniform(x_min, 1)
+        y_max = random.uniform(y_min, 1)
+        return Box2d(x_min, y_min, x_max, y_max)
+
+    list1 = [a_random_box2d() for _ in range(0, list1_len)]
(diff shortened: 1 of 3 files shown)


Renkai:
@eddyxu code updated, not sure if it is the right vectorized version because I'm new to vectorizing, the benchmark is being prepared but you can have a look at the code in advance if you'd like.

Renkai:
```
$ python python/benchmark/types/iou_benchmark.py 1000 1000 10

naive method cost 15.941821 seconds
vectorized method cost 0.360174 seconds
naive/vectorized 44.26144308028897
```
From my benchmark, the vectorized version is 44 times faster than the naive version for two parameters with length 1000.

python/benchmark/types/iou_benchmark.py
```diff
@@ -0,0 +1,61 @@
+import datetime
```
eddyxu, line 1: Good work! Could you add Apache license ?

Also run `isort` to put the imports into different sections?

python/rikai/types/geometry.py
```diff
@@ -281,6 +281,36 @@ def _area(arr: np.ndarray) -> np.array:
             0, arr[:, 3] - arr[:, 1]
         )
 
+    @staticmethod
+    def iou_matrix(box_list1: Union[Sequence[Box2d], np.ndarray],
```
eddyxu, line 285: lets rename them as `boxes1, boxes2` ? they are not necessary a `list` ;)

python/benchmark/types/iou_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+        x_max = random.uniform(x_min, 1)
+        y_max = random.uniform(y_min, 1)
+        return Box2d(x_min, y_min, x_max, y_max)
+
+    list1 = [a_random_box2d() for _ in range(0, list1_len)]
+    list2 = [a_random_box2d() for _ in range(0, list2_len)]
+
+    time0 = datetime.now()
+    count = 0
+    while count < times:
```
eddyxu, line 42: can we use [timeit](https://docs.python.org/3/library/timeit.html) here?

python/rikai/types/geometry.py
```diff
@@ -281,6 +281,36 @@ def _area(arr: np.ndarray) -> np.array:
...
+
+        xmin = np.maximum(box_list1[:, 0].reshape((row_count, -1)), box_list2[:, 0])
+        ymin = np.maximum(box_list1[:, 1].reshape((row_count, -1)), box_list2[:, 1])
+        xmax = np.minimum(box_list1[:, 2].reshape((row_count, -1)), box_list2[:, 2])
+        ymax = np.minimum(box_list1[:, 3].reshape((row_count, -1)), box_list2[:, 3])
+
+        inter_area = np.maximum(0, xmax - xmin) * np.maximum(
+            0, ymax - ymin)
+
+        iou_mat = inter_area / (area1 + area2 - inter_area)
```
eddyxu, line 310: do we need to handle `DivideByZero` here?

python/rikai/types/geometry.py
```diff
@@ -281,6 +281,36 @@ def _area(arr: np.ndarray) -> np.array:
             0, arr[:, 3] - arr[:, 1]
         )
 
+    @staticmethod
+    def iou_matrix(box_list1: Union[Sequence[Box2d], np.ndarray],
+                   box_list2: Union[Sequence[Box2d], np.ndarray]) -> np.ndarray:
+        """Compute intersection over union(IOU).
```
eddyxu, line 287: We use [numpy doc style](https://numpydoc.readthedocs.io/en/latest/format.html) in rikai. could you help to update the document to be aligned with the docstring style?

python/benchmark/types/iou_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    time1 = datetime.now()
+    count = 0
+    while count < times:
+        count += 1
+        Box2d.iou_matrix(list1, list2)
+
+    time2 = datetime.now()
+    naive_seconds = (time1 - time0).total_seconds()
+    vectorized_seconds = (time2 - time1).total_seconds()
+    print("naive method cost {} seconds".format(naive_seconds))
```
eddyxu, line 55: Nice, wonder what does the result look like

eddyxu:
Hey, @Renkai this is a nice work!

Can you help to report the numbers where the length of boxes are 10, 100, 1000? Also let us know your CPU model?

It is not mandate, but if possible, can we also get the same benchmark against MKL and openBlas (including MKL and openBlas version)?

For example, you can run

```python
>>> import numpy as np
>>> np.__config__.show()
```

to see what library is used.

python/rikai/types/geometry.py
eddyxu: Shall we just renames this as `ious`, and allow the function takes `Union[Sequence[Box2d], np.ndarray, Box2d]`?

python/rikai/types/geometry.py
Renkai: The default behavior is to print a runtime warning and gives `Nan` in the corresponding cell as a result, I believe it's an ok solution.

Renkai:
```
In [1]: import numpy as np

In [2]: np.__config__.show()
blas_mkl_info:
  NOT AVAILABLE
blis_info:
  NOT AVAILABLE
openblas_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/usr/local/lib']
    language = c
    define_macros = [('HAVE_CBLAS', None)]
blas_opt_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/usr/local/lib']
    language = c
    define_macros = [('HAVE_CBLAS', None)]
lapack_mkl_info:
  NOT AVAILABLE
openblas_lapack_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/usr/local/lib']
    language = c
    define_macros = [('HAVE_CBLAS', None)]
lapack_opt_info:
    libraries = ['openblas', 'openblas']
    library_dirs = ['/usr/local/lib']
    language = c
    define_macros = [('HAVE_CBLAS', None)]
```
Seems there is no `MKL` on my computer.

CPU Info
```
➜  ~ sysctl -n machdep.cpu.brand_string
Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
```

With the box list length of 10/100/1000, the result of time cost division is 1.79/19.42/42.57
```
(rikai-example) ➜  rikai git:(iou_two_lists) ✗ python python/benchmark/types/iou_benchmark.py 10 10 10
naive method cost 0.002192802000000049 seconds
vectorized method cost 0.0012242569999999509 seconds
naive/vectorized 1.7911288234415954
(rikai-example) ➜  rikai git:(iou_two_lists) ✗ python python/benchmark/types/iou_benchmark.py 100 100 10
naive method cost 0.163356304 seconds
vectorized method cost 0.008413267999999974 seconds
naive/vectorized 19.41651020744858
(rikai-example) ➜  rikai git:(iou_two_lists) ✗ python python/benchmark/types/iou_benchmark.py 1000 1000 10
naive method cost 16.681013051 seconds
vectorized method cost 0.391952670000002 seconds
naive/vectorized 42.55874325591382
```

eddyxu (approved):
Looks nice. Thanks @Renkai

PR #344, opened 2021-10-29
Support video format via DataSourceAPIv2

da-tubi:
Add a `video` format based on DataSourceAPIV2 for loading Video data.

``` scala
    val df = spark.read
      .format("video")
      .load("python/tests/assets/big_buck_bunny_short.mp4")
```

The equiv SQL snippet is not supported yet and can be supported later.
``` sql
select * from video.`s3:///path-to-video.mp4`
```

And a java udf is provided to extract the data part of the image:

``` python
from pyspark.sql.types import BinaryType
spark.udf.registerJavaFunction('image_data', "org.apache.spark.sql.rikai.expressions.ImageData", BinaryType())
```
In this way, we could invoke `ML_PREDICT` in this way:

``` sql
select ML_PREDICT(yolov5, image_data(image))
```

## TODO (not scheduled in this PR)
+ [ ] Video Options  (FFmpeg options) 
+ [ ] video magic format:
+ [ ] Now that javacv is introduced
    - [ ] video_to_images can be reimplemented in Scala side
    - [ ] documentation and deployment scripts (because simply `pip install rikai` does not work for video format related functionalities )
+ [ ] support `s3:///`, `https://`

``` sql
-- use case 1
select * from video.`path to video.mp4`
--- use case 2
create temporary view xxx options (
  path="xxx.mp4",
  option1=1,
  sampling=0.2
);

select ML_PREDICT(image_data(image)) from xxx
```

## Note
The adoption of OpenCV will make it a bit harder to use Rikai when the code path covers the OpenCV Java related code.


project/plugins.sbt
```diff
@@ -2,4 +2,5 @@ addSbtPlugin("com.geirsson" % "sbt-ci-release" % "1.5.6")
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "2.0.2")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.9.5")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.2")
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.15.0")
+addSbtPlugin("com.github.sbt" % "sbt-native-packager" % "1.9.6")
```
da-tubi, line 5: to generate the zip package of all jars

da-tubi:
```
sbt universal:packageBin
```
will create `target/universal/rikai-0.0.13-SNAPSHOT.zip`

To make it work with PySpark, we need to put the following two jars with PySpark.
```
org.bytedeco.ffmpeg-4.4-1.5.6-macosx-x86_64.jar
org.bytedeco.javacpp-1.5.6-macosx-x86_64.jar
```

Here is how I put these two jars to local PySpark:
```
cp *-macosx-x86_64.jar /Users/da/.pyenv/versions/3.8.10/envs/rikai-example/lib/python3.8/site-packages/pyspark/jars
```

eddyxu: Are there fallback-able java version for ffmpeg? Like if users forgot / dont know to install these two native packages, they can still run the code, just slower?

src/main/scala/ai/eto/rikai/sql/spark/datasources/VideoDataSourceV2.scala
```diff
@@ -0,0 +1,54 @@
...
+package ai.eto.rikai.sql.spark.datasources
+
+import org.apache.spark.sql.connector.catalog.Table
+import org.apache.spark.sql.execution.datasources.FileFormat
+import org.apache.spark.sql.execution.datasources.v2.FileDataSourceV2
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+
+class VideoDataSourceV2 extends FileDataSourceV2 {
+  override def shortName(): String = "video"
```
eddyxu, line 26: This means EACH video is ONE dataframe? Is this the desired use case?

cc @changhiskhan ?

project/plugins.sbt
da-tubi: no fallback-able java version now

src/main/scala/ai/eto/rikai/sql/spark/datasources/VideoDataSourceV2.scala
da-tubi: **Using the video format data source**, each video will be loaded as dataframe.

It is one of the use cases.

PR #346, opened 2021-11-06
MLflow backed MLCatalog

eddyxu:


Renkai:
take over it, will continue from the latest commit

changhiskhan:
@Renkai i think you may want this as well: https://github.com/eto-ai/rikai/pull/351
Currently we only set tags on the run, which makes it difficult to search models for a particular property. #351 gets you started with logging to the registered model and the version. To wrap it up and make it ready, please test it further and familiarize yourself with the associated Mlflow APIs. #351 and the mlflow catalog goes together

Issue #342, opened 2021-10-22
Upgrade minimal pytorch version to 1.9


eddyxu:
We can use torch 1.8's DataLoader and remove the DataLoader from `rikai.torch.data.DataLoader`

Renkai:
@eddyxu Do we need to keep these two tests(`test_load_dataset` and `test_load_dataset`) after we remove the class `rikai.torch.data.DataLoader`?


https://github.com/eto-ai/rikai/blob/88541f3140352161f9647c7f0caa5abfaac65442/python/tests/torch/test_data.py#L31

Renkai:
For new `Dataloader` after pytorch 1.9, is `test_torch_dataset` enough?

eddyxu:
Thanks for working on this, @Renkai .  Lets use `torch.data.DataLoader` in `test_load_dataset` to verify the compatibility between rikai dataset and torch DataLoader?

Renkai:
upgraded to 1.8, will create a ticket again when the newest Databricks Runtime is prepared for 1.9.

PR #349, opened 2021-11-10
downgrade pytorch to 1.8.0

Renkai:
Related issue: https://github.com/eto-ai/rikai/issues/342
Related pr: https://github.com/eto-ai/rikai/pull/347

Downgrade PyTorch version to align with Databricks platform, but we can still get rid of a customized `Dataloader`, we can upgrade it again when Databricks platform upgrades.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -24,7 +24,7 @@ dev = [
     "wheel",
 ]
 sklearn = ["scikit-learn"]
-torch = ["torch>=1.8.0", "torchvision>=0.11.0"]
+torch = ["torch>=1.8.1", "torchvision>=0.9.1"]
 jupyter = ["matplotlib", "jupyterlab"]
 aws = ["boto3", "botocore"]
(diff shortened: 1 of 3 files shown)


python/setup.py
```diff
@@ -24,7 +24,7 @@
     "wheel",
 ]
 sklearn = ["scikit-learn"]
-torch = ["torch>=1.9.0", "torchvision>=0.11.0"]
+torch = ["torch>=1.8.0", "torchvision>=0.11.0"]
```
eddyxu, line 27: do we need to change torchvision version as well?

da-tubi: See https://docs.databricks.com/release-notes/runtime/8.3ml.html

da-tubi: torchvision 0.9.1

Renkai: updated

darcy-shen (approved):
LGTM

PR #351, opened 2021-11-10
Set model tag

changhiskhan:
i'm having some local setup issue that's breaking unit test, putting this up for you to take look. will fix it tmr

### DIFF
--- a/python/rikai/spark/sql/codegen/mlflow_logger.py
+++ b/python/rikai/spark/sql/codegen/mlflow_logger.py
@@ -129,9 +129,23 @@ class MlflowLogger:
                     "will not be populated to MLflow"
                 )
         mlflow.set_tags(tags)
-        model_registry_client = MlflowClient()._get_registry_client()
-        for key, value in tags.items():
-            model_registry_client.set_registered_model_tag(registered_model_name, key, value)
+        if registered_model_name is not None:
+            # if we're creating a model registry entry, we also want to set the tags on the model version
+            # and model to enable search
+            c = MlflowClient()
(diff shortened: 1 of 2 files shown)


python/rikai/spark/sql/codegen/mlflow_logger.py
```diff
@@ -128,6 +129,9 @@ def log_model(
                     "will not be populated to MLflow"
                 )
         mlflow.set_tags(tags)
+        model_registry_client = MlflowClient()._get_registry_client()
+        for key, value in tags.items():
+            model_registry_client.set_registered_model_tag(registered_model_name, key, value)
```
eddyxu, line 134: Question here, should the tag attach to a register model or a register model version?

Can user change some of these flags from different versions

changhiskhan: this is the issue I was complaining to you about. When you call mlflow log_model, it creates a new version but doesn't return the version id to you. My thought is that we probably want to log properties to both the model and the version. When you create a new version, the latest version's properties should override the ones on the registered model. The intended workflow is that most of the time, users will search registered models and then just use the latest. And if there's some debugging workflow, they can click through historical versions and try to find any changes.

changhiskhan:
@Renkai do you have bandwidth to tackle this one next?

Renkai:
@changhiskhan I think I would have, but the description on this PR is too simplified that I can't get the complete goal of it. Maybe we could discuss it in detail in the next community meeting.

Renkai:
take over this PR

PR #352, opened 2021-11-11
Catalog constructor takes in a SparkConf

eddyxu:
Allow customized Catalog behavior at runtime


changhiskhan (approved):
so the catalog config is set for the duration of the session when the first command touching the catalog is invoked right?

eddyxu:
Yea, so right now the configure should be set via a cluster deployment. One cluster only has one model catalog.

PR #358, opened 2021-11-19
udf: array_box2d, inspect_bbox

da-tubi:
Why we need `array_box2d`, because:

``` sql
select inspect_bbox(image, transform(pred.boxes, x -> array_box2d(x)), transform(pred.label_id, i -> string(i)))
```

```
Error in SQL statement: UnsupportedOperationException: Cannot generate code for expression: inspect_bbox(input[4, image, true], transform(input[3, struct<boxes:array<array<float>>,label_ids:array<int>>, true].boxes, lambdafunction(box2d(lambda arr#393), lambda arr#393, false)), transform(input[3, struct<boxes:array<array<float>>,label_ids:array<int>>, true].label_ids, lambdafunction(cast(lambda i#394 as string), lambda i#394, false)))
```

### DIFF
--- a/python/rikai/spark/functions/vision.py
+++ b/python/rikai/spark/functions/vision.py
@@ -22,6 +22,7 @@ from typing import List, Optional, Union
 
 # Third Party
 import numpy as np
+from PIL import ImageDraw
 from pyspark.sql.functions import udf
 from pyspark.sql.types import (
     ArrayType,
@@ -50,6 +51,7 @@ __all__ = [
     "crop",
     "to_image",
     "image_copy",
+    "inspect_bbox",
     "numpy_to_image",
     "video_to_images",
     "spectrogram_image",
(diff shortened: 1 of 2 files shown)


python/rikai/spark/functions/vision.py
```diff
@@ -453,3 +455,36 @@ def crop(img: Image, box: Union[Box2d, List[Box2d]]):
 
     """
     return img.crop(box)
+
+
+@udf(returnType=ImageType())
+def inspect_bbox(img: Image, boxes: List[Box2d], labels: List[str]):
+    """Draw bounding boxes on the image, and the returns the inpected image
```
eddyxu, line 462: Do we need to put this into a SQL as fixed open API from Rikai?

For example, how should we define color, font, texture, transparency, location of texts to boxes? Or do we want to?

python/rikai/spark/functions/vision.py
```diff
@@ -453,3 +455,36 @@ def crop(img: Image, box: Union[Box2d, List[Box2d]]):
...
+    >>> spark.sql("SELECT inspect_bbox(img, boxes, labels) as bbox FROM detections")
+
+    """
+    assert len(boxes) == len(labels), "length of boxes and labeles must match"
+    pil_image = img.to_pil()
+    draw = ImageDraw.Draw(pil_image)
+    for (bbox, label) in zip(boxes, labels):
+        draw.rectangle(bbox.to_numpy().tolist(), outline="green", width=2)
+        draw.text([bbox.xmin + 5, bbox.ymin - 10], label, fill="red")
+    return Image.from_pil(pil_image)
```
eddyxu, line 490: So this means that all images are loaded into memory after this point?

python/rikai/spark/functions/vision.py
darcy-shen: I think this simple bounding boxes drawer udf is sufficient for Rikai.

UDF can be easily adapted by nature. I guess there are professional pip packages to draw bounding boxes.

darcy-shen: For example: https://github.com/varunagrawal/bbox

Personally, I'm in favor of keeping only the necessary dependencies. For advanced usage, we could put it in the contrib folder.

python/rikai/spark/functions/vision.py
darcy-shen: Yes. To draw on images, it must be loaded into memory.

This UDF should not be used during batch processing if we care about the storage/computing cost. It is for visualization purpose.

changhiskhan: if we need viz specific functionality, i'm not sure a UDF is the right place to put it. Instead, I think we need the following:
1. Each annotation type should know how to draw itself on a given image
2. A function that takes a single image and a number of annotations and outputs an annotated image (using 1). I think this might be something that we have already (I vaguely remember seeing a PR from @eddyxu). 
3. A function that takes in a DataFrame and it calls 2 on the first N rows. Optionally, add some layout parameters.

changhiskhan (changes requested):

changhiskhan:
also, @da-tubi why are you requesting reviews from yourself LOL

da-tubi:
Because I am using @darcy-shen to reply some review comments.

Well, at the very beginning, I missed the email invitation by @da-tubi and finally accepted the invitation by @darcy-shen .

Could you remove @darcy-shen and invite @da-tubi again instead. I'd like to separate work coding from life coding and keep a balance between life and work.

PR #359, opened 2021-11-19
Register all rikai UDFs using init

da-tubi:


python/rikai/spark/functions/__init__.py
```diff
@@ -15,6 +15,22 @@
...
 from rikai.spark.functions.vision import *
+
+__all__ = ["init"]
+
+
+def init(spark: SparkSession):
+    """Register all rikai UDFs"""
+    from rikai.spark.functions import geometry, io, vision
+
+    for name in geometry.__all__:
```
eddyxu, line 31: Shall we make sure that these exported methods are actually spark udf?

darcy-shen: I do not think so. Because the set of Rikai built-in is a finite set!

Let it crash when it is not actually a spark udf.

darcy-shen: I've checked the source code of PySpark. There are no ready-to-use functions to check if a python function is a valid PySpark udfs.

This pr assumes the public methods defined in `__all__` for these this python files should be valid pyspark udf.

If one of them is not valid, it will be found during Github Action.

python/rikai/spark/functions/__init__.py
```diff
@@ -15,6 +15,22 @@
...
+        spark.udf.register(name, getattr(geometry, name))
+    for name in io.__all__:
+        spark.udf.register(name, getattr(io, name))
+    for name in vision.__all__:
+        spark.udf.register(name, getattr(vision, name))
```
changhiskhan, line 36: what happens if you don't have all of the optional dependencies installed (like ffmpeg-python for example)?

python/rikai/spark/functions/__init__.py
changhiskhan: i think it's likely that we'll have a larger set of UDFs than what we want to always register at init. "explicit is better than implicit" after all.

changhiskhan (commented):
thanks for doing this. some minor changes/questions

python/rikai/spark/functions/__init__.py
eddyxu: I guess what I meant is that, this PR prevents `geometry` and etc modules to export any non `udf` functions/classes in `__all__`.

da-tubi: how about add a new global variable `__all_udfs__` for this case?

changhiskhan: How about:
1. we create a `RikaiUdfRegistry` singleton class that contains an internal dict of {name -> udf func}
2. we create a `@rikai_udf` decorator so that it creates a spark udf and also registers the function in the registry
3. within this `init` function, just call `registry.register_all()`

@da-tubi @eddyxu wdyt?

python/rikai/spark/functions/__init__.py
darcy-shen: throw error during runtime

changhiskhan (approved):
ok to merge, but one suggested improvement is to make the list the actual functions for easier linting:

e.g., `all_geo_udfs = [geometry.area, ...]`

but as long as this is unit tested there's not much risk

Issue #357, opened 2021-11-19
Databricks Notebook support for Image.display 


darcy-shen:

changhiskhan:
did you find a good way to do it? I searched around and it wasn't clear what API they were tapping into

darcy-shen:
Here is my workaround
``` python
import matplotlib.pyplot as plt

def imshow(image):
  plt.imshow(image.to_pil())
  plt.show()


image = (
  spark.sql("select to_image('https://github.com/da-tubi/rikai-example/raw/main/img/lena.png') as image")
     .first()
     .image
)
imshow(image)
```

darcy-shen:
https://community.databricks.com/s/question/0D53f00001HKHrwCAH/how-to-show-an-image-in-a-notebook-using-html
```
from IPython.display import Image 
from IPython.core.display import HTML 
Image(url= "https://docs.databricks.com/_images/landing-aws.png")
```

> For small images, another cheeky alternative is to base64-encode the image bytes and construct a data:/ URI out of it, and make a hyperlink to that. It is 'inline' in the notebook then, but, wouldn't be a good idea for big images.

darcy-shen:
```
import rikai
import base64

def display(image: rikai.types.vision.Image):
    from IPython.core.display import Image
    
    if not image.is_embedded and image.uri.startswith("http"):
        return Image(url = image.uri)
    else:
        with image.open() as fobj:
            encoded = base64.b64encode(fobj.read()).decode("utf-8")
            url = f"data:image;base64,{encoded}"
            return Image(url = url)
```

Tested in databricks notebook:

### Test Case 1

![image](https://user-images.githubusercontent.com/1267865/142799023-2dd3267f-75c5-48d4-b6bd-e9d7e3647f8c.png)

### Test Case 2

![image](https://user-images.githubusercontent.com/1267865/142799095-b97059fd-148b-4c85-8b38-988b6db5cac8.png)


### Test Case 3
![image](https://user-images.githubusercontent.com/1267865/142799180-5ab7db8d-977a-43a5-8812-e344aeb43377.png)

darcy-shen:
closed by #360

PR #360, opened 2021-11-22
Tune image display on Databricks

da-tubi:
Tested on Databricks: see #357 

Here is the test cases in the jupyter notebook:

image

image


changhiskhan (approved):
nice one!
Can you add a comment that this is required for databricks notebook?

eddyxu:
TIL ! Nice work!

Issue #354, opened 2021-11-18
Enhance the customization of the generated PyTorch pandas UDF


darcy-shen:
## Current Status
+ pre_processors
+ post_processors

## Limitation
Intermediates produced by pre_processors can not be passed to post_processors.

### Extra parameters (inputs) for the post_processors
See #364 

Besides predictions, we also pass the inputs of the model to post_processors.

#### Case 1: needs to be logged to MLflow
See #361

#### Case 2: use the Module Wrapper in the generated pandas udf, no need to log the model
```
create or replace model
preprocessor 'x.y.z.pre_procecssing'
postprocessor 'x.y.z.post_processing'
wrapper 'x.y.z.WrappedModule'
...
```

darcy-shen:
See #361 

Sometimes, for example the yolov5 model, to make it work with Rikai. We have to create a wrapper model to separate these three parts: pre_processing/model.forward/post_processing.

Simply adding extra parameters to pre_processor/post_processor is not enough for the yolov5 model.

PR #361, opened 2021-11-22
Yolov5 contrib

da-tubi:
How to test the yolov5 contrib: see github action

<del>This PR is nearly complete, I still need to do some clean-ups. I'd like to receive some reviews first.</del>

## Changes
+ pre_processor api new design
+ yolov5 contrib using the new pre_processor api


python/rikai/contrib/torch/__init__.py
```diff
@@ -11,3 +11,16 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+
+from abc import ABC, abstractmethod
+from typing import Any, Dict
+
+
+class RikaiModule(ABC):
```
eddyxu, line 19: Do you intent to store this model into model registry? Will this interface break other workload, i.e., model deployment in non-rikai enrionment?

contrib/yolov5/setup.py
```diff
@@ -0,0 +1,22 @@
...
+
+setup(
+    name="rikai-yolov5",
+    version="0.0.1",
+    license="Apache License, Version 2.0",
+    author="Rikai authors",
+    author_email="rikai-dev@eto.ai",
+    url="https://github.com/eto-ai/rikai",
+    python_requires=">=3.7",
+    install_requires=["rikai", "yolov5==5.0.8"],
```
eddyxu, line 11: so do we want to maintain a new version of this `contrib` package for each yolov5 new version>

contrib/yolov5/rikai/contrib/yolov5/model.py
```diff
@@ -0,0 +1,79 @@
...
+    make_divisible,
+    non_max_suppression,
+    scale_coords,
+)
+from yolov5.utils.torch_utils import time_sync
+
+from rikai.contrib.torch import RikaiModule
+
+
+class RikaiYolov5Model(nn.Module, RikaiModule):
```
eddyxu, line 17: Lets just rename it to `Yolov5`?

And how does this work with traning, deployment and etc?

python/rikai/contrib/torch/__init__.py
da-tubi: > Do you intent to store this model into model registry?
Yes

> Will this interface break other workload
No

Because the model logged for Rikai (batch processing purpose) can be separated from the model for serving and other use cases.

contrib/yolov5/setup.py
da-tubi: Updated, restrict it in [5.0.0, 6.0.0), because I have not tested it for 6.x.

contrib/yolov5/rikai/contrib/yolov5/model.py
da-tubi: > Lets just rename it to Yolov5?

Renamed.

> And how does this work with traning, deployment and etc?

This interface adds an extra parameter: option, in this way, we could pass more options via

``` sql
create or replace model xxx
options(
 op1='x',
 op2=0.3
)
```

python/rikai/contrib/torch/__init__.py
da-tubi: > PickleError: ScriptModules cannot be deepcopied using copy.deepcopy or saved using torch.save. Mixed serialization of script and non-script modules is not supported. For purely script modules use my_script_module.save(<filename>) instead.

```
import torch.nn as nn
import torch

class WrappedModel(nn.Module):
    def __init__(self, model):
        super().__init__()
        self.model = model.eval()

    @torch.no_grad()
    def forward(self, imgs):
        preds = self.model(imgs)
        return (preds, imgs)
```

I tried to log the wrapped model to MLflow, but failed, because it is a torch script model.

da-tubi: There are ways to use the wrapped model without logging it to MLflow (or model store).

As discussed in https://github.com/eto-ai/rikai/issues/354

da-tubi:
@eddyxu  A much more flexible new design inspired by @changhiskhan at #364 

This new design is flexible enough to solve #367

PR #364, opened 2021-11-23
Enhanced post processor with inputs passed

da-tubi:
One of the various way to Enhance the customization of the generated PyTorch pandas UDF

See #354

### DIFF
--- a/python/rikai/spark/sql/codegen/pytorch.py
+++ b/python/rikai/spark/sql/codegen/pytorch.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 
 import os
+from inspect import signature
 from typing import Iterator
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


python/rikai/spark/sql/codegen/pytorch.py
```diff
@@ -74,7 +75,9 @@ def torch_inference_udf(
                     if isinstance(batch, torch.Tensor):
                         batch = batch.to(device)
                     predictions = model(batch)
-                    if spec.post_processing:
+                    if _number_of_params(spec.post_processing) == 2:
+                        predictions = spec.post_processing(batch, predictions)
```
changhiskhan, line 79: do we want to have these optional parameters that we try to figure out whether it's present? I'm wondering whether it would be better to be more explicit here? Options:
1. Just have predictions and model be required arguments
2. declare the post processing interface in the model config
3. something else?

da-tubi: > Just have predictions and model be required arguments

Typo? Just have predictions and batch (model input) be required arguments?

da-tubi: > Just have predictions and model be required arguments

Cool idea

PR #369, opened 2021-11-25
MLflow backed MLCatalog

Renkai:
This patch is continue from https://github.com/eto-ai/rikai/pull/346

* This patch contains a slightly hacked Mlflow Java client, because the official java client doesn't support `DELETE with body` in HTTP, which Python client support and needed in rest API https://www.mlflow.org/docs/1.21.0/rest-api.html#delete-registeredmodel
* Mlflow has more concepts than our current `trait Catalog`, like versions and stages, if `Mlflow` would be the de facto standard in the future, maybe we need to add those in trait. Currently, it only picks the latest version in all stages.
* Filter by `ArtifactPathKey` temporarily disabled because we need apply https://github.com/eto-ai/rikai/pull/351 first.


src/main/java/org/mlflow/tracking/RiMlflowClient.java
```diff
@@ -0,0 +1,73 @@
...
+import org.mlflow.tracking.creds.*;
+
+import java.net.URI;
+
+
+/**
+ * Client to an MLflow Tracking Sever.
+ * Originally from mlflow client 1.21.0
+ */
+public class RiMlflowClient extends MlflowClient {
```
eddyxu, line 27: What does `Ri` mean, `Rikai`?  If so lets just name it as `RikaiMlflowClient` and make it a package visible class?  This is not design as user-facing API right? 

Should we actually provide a patch to `mlflow` java SDK as well?

src/main/java/org/mlflow/tracking/RiMlflowHttpCaller.java
```diff
@@ -0,0 +1,246 @@
...
+import org.apache.http.ssl.SSLContextBuilder;
+import org.apache.http.util.EntityUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.mlflow.tracking.creds.MlflowHostCreds;
+import org.mlflow.tracking.creds.MlflowHostCredsProvider;
+
+// Originally from mlflow client 1.21.0
+class RiMlflowHttpCaller {
```
eddyxu, line 51: These are a lot of code ;) 

Can we add a TODO plan that eventually we add DELETE to mlflow, and remove these code?

src/main/scala/ai/eto/rikai/sql/model/Catalog.scala
```diff
@@ -75,6 +75,7 @@ object Catalog {
           .asInstanceOf[Catalog]
       )
     }
+    println(s"catalog get ${catalog.get.getClass}")
```
eddyxu, line 78: remove println or change it to `logger.debug`?

src/main/scala/ai/eto/rikai/sql/model/mlflow/MlflowCatalog.scala
```diff
@@ -0,0 +1,142 @@
...
+/** Use MLflow as a persisted backend for Model Catalog
+  */
+class MlflowCatalog(val conf: SparkConf) extends Catalog {
+
+  private val mlflowClient = new MlflowClientExt(conf.get(TrackingUriKey))
+
+  /** Create a ML Model that can be used in SQL ML in the current database.
+    */
+  override def createModel(model: Model): Model = {
+    val name = mlflowClient.createModel(
```
eddyxu, line 37: How does it work with mlflow.

Does it mean that for each registered model, it will create a separate model in MLflow?

eddyxu (commented):
Looks nice overall.  Have a few questions

src/test/scala/ai/eto/rikai/sql/model/mlflow/SparkSessionWithMlflow.scala
```diff
@@ -0,0 +1,75 @@
...
+import ai.eto.rikai.sql.model.{Catalog, Registry}
+import org.apache.logging.log4j.scala.Logging
+import org.apache.spark.sql.SparkSession
+import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach, Suite}
+
+import scala.collection.JavaConverters._
+
+trait SparkSessionWithMlflow
+    extends Logging
+    with BeforeAndAfterEach
```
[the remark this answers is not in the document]
Renkai: Not necessary here, we can add this in sub classes, removed

src/main/java/org/mlflow/tracking/RiMlflowClient.java
Renkai: Renamed it to `RikaiMlflowClient`, it's actually a patch of `MlflowClient`, so I keep the same visible level and package name with `RikaiMlflowClient`.
It's a good idea to provide a patch to the mlflow official repo, I will try it. And we can remove this file after patch merged by mlflow official.

src/main/java/org/mlflow/tracking/RiMlflowHttpCaller.java
Renkai: Like `RikaiMlflowClient`, we can remove this whole file after the official mlflow SDK provide the delete feature. Before that, I think keeping this file is convenient for test reasons. We can also add a TODO to plan to remove this file after mlflow officially gives this feature.

src/main/scala/ai/eto/rikai/sql/model/Catalog.scala
Renkai: updated

src/main/scala/ai/eto/rikai/sql/model/mlflow/MlflowCatalog.scala
Renkai: Yes, I think so, each line in `show models` corresponds to the newest version of each model in Mlflow.

darcy-shen:
Tips: use `make fix` to do formatting.

Renkai:
I found some errors in the tests can be fixed by a longer limit time in my local machine. And CI shows the different error logs from my local computer, adding more commits for testing.

.github/workflows/python.yml
```diff
@@ -44,9 +48,11 @@ jobs:
     - name: Run python tests
       working-directory: python
       run: |
-        pytest -v --durations=10 --ignore=tests/parquet/internal
+        sudo chmod 777 /var/lib #mlflow local temp dir need it
```
da-tubi, line 51: better not to use `/var/lib`

da-tubi: why not put it under /tmp/

.github/workflows/python.yml
```diff
@@ -28,6 +28,10 @@ jobs:
     - name: sbt install
       run: |
         sbt publishLocal
+    - name: Start docker-compose
```
da-tubi, line 31: using mlflow via os multiple process is sufficient, I think

python/tests/conftest.py
```diff
@@ -21,20 +21,116 @@
...
+    mlflow.set_tracking_uri(tracking_uri)
+    experiment_id = mlflow.create_experiment("rikai-test", str(tmp_path))
+    # simpliest
+    with mlflow.start_run(experiment_id=experiment_id):
+        mlflow.log_param("optimizer", "Adam")
+        # Fake training loop
+        model = torch.load(resnet_model_uri)
+        artifact_path = "model"
+
+        schema = (
```
da-tubi, line 53: nit: can be imported

python/tests/conftest.py
```diff
@@ -21,20 +21,116 @@
...
+
+@pytest.fixture(scope="module")
+def spark(mlflow_tracking_uri: str) -> SparkSession:
+    print(f"ml flow tracking uri for spark: ${mlflow_tracking_uri}")
     rikai_version = get_default_jar_version(use_snapshot=True)
     hadoop_version = "3.2.0"  # TODO(lei): get hadoop version
+    # Avoid reused session polluting configs
+    active_session = SparkSession.getActiveSession()
+    if active_session:
+        print("active session stopped, will restart")
```
da-tubi, line 132: nit: print

python/tests/spark/sql/codegen/conftest.py
```diff
@@ -0,0 +1,156 @@
...
+    )
+    # simpliest
+    with mlflow.start_run(experiment_id=experiment_id):
+        mlflow.log_param("optimizer", "Adam")
+        # Fake training loop
+        print("resnet model uri", resnet_model_uri)
+        model = torch.load(resnet_model_uri)
+        artifact_path = "model"
+
+        schema = (
```
da-tubi, line 50: nit: can be imported

python/tests/spark/sql/codegen/test_mlflow_catalog.py
```diff
@@ -0,0 +1,30 @@
...
+    spark = spark_with_mlflow
+    x = np.array([[1, 1], [1, 2], [2, 2], [2, 3]])
+    y = np.dot(x, np.array([1, 2])) + 3
+    model = LinearRegression()
+    with mlflow.start_run():
+        model.fit(x, y)
+
+    tracking_uri = mlflow.get_tracking_uri()
+    print("Tracking URI", tracking_uri)
+    spark.sql("SHOW MODELS").show()
```
da-tubi, line 30: nit: better to use assertion

src/main/scala/ai/eto/rikai/sql/model/Catalog.scala
```diff
@@ -52,6 +53,7 @@ trait Catalog {
 }
 
 object Catalog {
+  private val logger = LoggerFactory.getLogger(this.getClass)
```
da-tubi, line 56: use LazyLogging from scala-logging

src/main/scala/ai/eto/rikai/sql/model/mlflow/MlflowCatalog.scala
```diff
@@ -0,0 +1,142 @@
...
+      model.flavor
+    )
+  }
+
+  /** Return a list of models available for all Sessions */
+  override def listModels(): Seq[Model] = {
+    val response = mlflowClient.searchRegisteredModels()
+    response.getRegisteredModelsList.asScala
+      .map(model => {
+        model.getLatestVersionsCount match {
```
da-tubi, line 54: simplify match to if?

src/main/scala/ai/eto/rikai/sql/model/mlflow/MlflowCatalog.scala
```diff
@@ -0,0 +1,142 @@
...
+object MlflowCatalog {
+
+  val TrackingUriKey = "rikai.sql.ml.registry.mlflow.tracking_uri"
+
+  val ArtifactPathKey = "rikai.model.artifact_path"
+  val ModelFlavorKey = "rikai.model.flavor"
+  val OutputSchemaKey = "rikai.output.schema"
+  val SpecVersionKey = "rikai.spec.version"
+  val PreProcessingKey = "rikai.transforms.pre"
+  val PostProcessingKey = "rikai.transforms.post"
```
da-tubi, line 139: for string literal, how about POST_PROCESSING_KEY

.github/Dockerfile.mlflow
```diff
@@ -0,0 +1,8 @@
+FROM python:3.9-slim
+
+RUN pip install mlflow==1.20.2
+
+CMD ["mlflow", "server", "--host", "0.0.0.0", "--workers", "1", \
+     "--backend-store-uri", "sqlite:///var/lib/mlflow.db", \
```
da-tubi, line 6: use /tmp to remove chmod on /var/lib below

python/tests/conftest.py
```diff
@@ -21,20 +21,116 @@
 from urllib.parse import urlparse
 
 # Third Party
+import mlflow
 import pytest
 import torch
 import torchvision
+from mlflow.tracking import MlflowClient
```
da-tubi, line 28: isort

python/tests/spark/sql/codegen/conftest.py
```diff
@@ -0,0 +1,156 @@
...
+from __future__ import annotations
+
+import datetime
+import os
+
+import mlflow
+import pytest
+import torch
+import torchvision
+from mlflow.tracking import MlflowClient
```
da-tubi, line 24: isort (i for import)

python/tests/conftest.py
Renkai: `make fix`, which contains `isort`, doesn't fix it automatically, is that because `isort` is not properly set?

.github/workflows/python.yml
Renkai: good idea, updated

python/tests/conftest.py
Renkai: I don't understand your comment here.

python/tests/conftest.py
Renkai: It's in test code base, so I think print is ok?

src/main/scala/ai/eto/rikai/sql/model/mlflow/MlflowCatalog.scala
Renkai: updated

python/tests/conftest.py
da-tubi:
``` python
from rikai.contrib.torch.transforms.fasterrcnn_resnet50_fpn import OUTPUT_SCHEMA
```

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -221,6 +221,7 @@ def resolve(self, raw_spec):
         name = raw_spec.getName()
         uri = raw_spec.getUri()
         logger.info(f"Resolving model {name} from {uri}")
+        logger.info(f"tracking uri: {self.mlflow_tracking_uri}")
```
da-tubi, line 224: How about "Using tracking uri: "

src/main/scala/ai/eto/rikai/sql/model/mlflow/MlflowClientExt.scala
```diff
@@ -0,0 +1,117 @@
...
+  private[mlflow] def searchRegisteredModels()
+      : SearchRegisteredModels.Response = {
+    val payload = client.sendGet("registered-models/search")
+    val builder = SearchRegisteredModels.Response.newBuilder()
+    logger.debug(s"Search Register Model response: ${payload}")
+    MlflowClientExt.merge(payload, builder)
+    builder.build()
+  }
+
+  private[mlflow] def getModel(name: String) = {
```
da-tubi, line 48: nit: type annotation

src/main/scala/ai/eto/rikai/sql/model/mlflow/MlflowCatalog.scala
da-tubi: Can be simplified using collect.

``` scala
    val response = mlflowClient.searchRegisteredModels()
    response.getRegisteredModelsList.asScala.collect {
      case model if model.getLatestVersionsCount != 0 =>
        val latestVersion = model.getLatestVersions(0)
        val tagsMap = latestVersion.getTagsList.asScala
          .map(t => t.getKey -> t.getValue)
          .toMap
        val name = model.getName
        val flavor = tagsMap.getOrElse(MODEL_FLAVOR_KEY, "")
        new SparkUDFModel(
          name,
          s"mlflow://$name",
          "<anonymous>",
          flavor
        )
    }
```

da-tubi: Did not test the code snippet because it requires a MLflow service.

python/tests/conftest.py
da-tubi: I don't know. How about isort it manually?

.github/workflows/python.yml
Renkai: That sounds like a good idea, I may test it with another PR.

da-tubi (approved):
LGTM

darcy-shen (approved):

da-tubi:
R API has the impl of delete registered model: https://github.com/mlflow/mlflow/pull/4483

PR #376, opened 2021-12-06
rikai-yolov5 and flavor customization support

### LINKED ISSUE #374: Customizing generate_udf
For Rikai 0.0.12, generate_udf is fixed for flavor. We have two flavors (torch, sklearn), and the corresponding two built-in generate_udfs.

https://github.com/eto-ai/rikai/blob/6971f676cc9e89aa4eabaebce94c7ea04ac765f1/python/rikai/spark/sql/codegen/base.py#L171

Can we provide a way for user to customize their own generate_udf
like if flavor is tubi.rikai.xxx.yyy.generate_udf, we could import the tubi-customized generate_udf

da-tubi:
closes #374 

## What
1. Provide ways to customize the flavor and the corresponding UDF generator.
    + Doc: will create another pull request on documentation
2. rikai-yolov5: A live use case of customizing the pytorch flavor
    + Doc: https://github.com/da-tubi/rikai/blob/feature/sklearn_contrib/contrib/yolov5/REAMDE.md


changhiskhan:
based on the discussion today, let's make sure we document this for users

changhiskhan (approved):
For the documentation can you add a blurb in the beginning about why this is needed (as opposed to just a regular pytorch flavored model). Also make a note that this is a pytorch model.

contrib/yolov5/REAMDE.md
```diff
@@ -0,0 +1,49 @@
+# Yolov5 support for Rikai
+## Usage
```
eddyxu, line 2: Could you have a follow up ticket to mention:

1. this is pytorch yolov5
2. and which implementation of yolov5

eddyxu (approved):
LGTM overall

PR #379, opened 2021-12-07
[PARQUET] Set default scheme

changhiskhan:
This is a convenience feature so that users can omit
the scheme of the dataset uri if they will always be
the same.


eddyxu (approved):
LGTM

PR #382, opened 2021-12-10
Segmentation Semantic Type

eddyxu:
Add `Mask` type, which presents a mask layer over a 2D image.  It can be used as representing segmentation in object detection task.


eddyxu:
@da-tubi would appreciate some inputs from you. what should be the good way to pass array into struct type?

I've got this error:

```
[info] - test create mask data *** FAILED ***
[info]   java.lang.RuntimeException: Error while encoding: java.lang.ClassCastException: class org.apache.spark.sql.catalyst.expressions.UnsafeArrayData cannot be cast to class org.apache.spark.sql.catalyst.InternalRow (org.apache.spark.sql.catalyst.expressions.UnsafeArrayData and org.apache.spark.sql.catalyst.InternalRow are in unnamed module of loader 'app')
[info] knownnotnull(assertnotnull(input[0, scala.Tuple2, true]))._1 AS _1#0
[info] newInstance(class org.apache.spark.sql.rikai.MaskType).serialize AS _2#1
[info]   at org.apache.spark.sql.catalyst.encoders.ExpressionEncoder$Serializer.apply(ExpressionEncoder.scala:213)
```

from this line https://github.com/eto-ai/rikai/pull/382/files#diff-359425d03a775e21651d2f9f81053e902305a47273ce377306bd9c00419e3f7aR86

```
row.update(4, UnsafeArrayData.fromPrimitiveArray(m.rle.get))
```

The [upstream spark code](https://github.com/apache/spark/blob/739dacd38ef7f160d77336baa076eefd016e2c40/mllib/src/main/scala/org/apache/spark/ml/linalg/MatrixUDT.scala#L55-L58)  seems to be the same to me.

```

```

contrib/coco/scripts/make_dataset.py
```diff
@@ -0,0 +1,64 @@
...
+    annotations = []
+    polygon_images = 0
+    rle_images = 0
+    for image_id in islice(coco.imgs, 0, 200):
+        img = coco.loadImgs(image_id)
+        anns = coco.loadAnns(coco.getAnnIds(imgIds=image_id))
+        if any([ann["iscrowd"] for ann in anns]):
+            rle_images += 1
+            if rle_images > 10:
+                break
```
changhiskhan, line 42: nit: though most of the time this , this is dependent on your sampling distribution. How about changing this to continue and add an `if rle_images > 10 and polygon_images > 10: break` below this if-else?

contrib/coco/tests/test_convert.py
```diff
@@ -0,0 +1,44 @@
...
+        img = coco.loadImgs(image_id)[0]
+        # print(img)
+        height, width = img["height"], img["width"]
+        ann_ids = coco.getAnnIds(imgIds=image_id)
+        for ann in coco.loadAnns(ann_ids):
+            if ann["iscrowd"] == 0:
+                mask = Mask.from_polygon(
+                    ann["segmentation"], height=height, width=width
+                )
+                # TODO: currently, it has disparity of the conversion between
```
changhiskhan, line 38: why is this disparity happening?
is this not important to address right now?

python/rikai/spark/types/geometry.py
```diff
@@ -148,3 +152,73 @@ def deserialize(self, datum: Row) -> "Box3d":
 
     def simpleString(self) -> str:
         return "Box3dType"
+
+
+class MaskType(UserDefinedType):
```
changhiskhan, line 157: add to __all__?

python/rikai/types/geometry.py
```diff
@@ -440,3 +448,207 @@ def to_dict(self) -> dict:
...
+        width: Optional[int] = None,
+        mask_type: Mask.Type = Type.POLYGON,
+    ):
+        self.type = mask_type
+        self.data = data
+        if self.type == Mask.Type.MASK:
+            self.height = data.shape[0]
+            self.width = data.shape[1]
+        else:
+            assert (
```
changhiskhan, line 524: nit: asserts get removed if optimized mode is turned on. So this is likely to result in surprising behavior based on user configurations

python/rikai/types/geometry.py
```diff
@@ -440,3 +448,207 @@ def to_dict(self) -> dict:
...
+            )
+
+    """
+
+    __UDT__ = MaskType()
+
+    class Type(Enum):
+        """Mask type."""
+
+        MASK = 0
```
changhiskhan, line 506: this is a bit confusing that you can have a Mask.Type be MASK

python/rikai/types/geometry.py
```diff
@@ -440,3 +448,207 @@ def to_dict(self) -> dict:
...
+    def to_numpy(self) -> np.ndarray:
+        return self.to_mask()
+
+    def to_dict(self) -> dict:
+        d = {
+            "type": self.type.value,
+            "width": self.width,
+            "height": self.height,
+        }
+        if self.type == Mask.Type.RLE:
```
changhiskhan, line 638: since the type is there, why not just name the key "data"? otherwise you need to interpret the type whenever you need to know what key to look at. Also it makes it easier to use Mask constructor with **kwargs?

python/rikai/types/geometry.py
```diff
@@ -440,3 +448,207 @@ def to_dict(self) -> dict:
...
+            d["rle"] = self.data
```
changhiskhan, line 639: so why call to_mask if it's COCO_RLE but not regular RLE?

python/rikai/types/rle.py
```diff
@@ -0,0 +1,65 @@
...
+#  limitations under the License.
+
+from __future__ import annotations
+
+from typing import Tuple
+
+import numpy as np
+
+
+def encode(arr: np.ndarray) -> np.array:
```
changhiskhan, line 22: Do we need to support multi-class semantic segmentation? I think binary is fine for now but am wondering whether we need to leave room in the design for it?

python/rikai/types/geometry.py
eddyxu: What about naming it to `Mask.Type.RAW`? or sth else?
So basically, polygon and RLE are compressed mask, and this `Mask.Type.MASK` is the original 2d matrix.

python/rikai/types/rle.py
eddyxu: So in the coco dataset at least, a labelled segmentation is a combination of `(label:str, mask:nd.array/RLE)`. So it uses label id to identify different classes.

contrib/coco/tests/test_convert.py
eddyxu: So, in [coco](https://github.com/cocodataset/cocoapi/blob/8c9bcc3cf640524c4c20a9c40e89cb6a2f2fa0e9/common/maskApi.c#L162),  it uses its own implementation to fill the polygon, which is different to `PIL / OpenCV`. 

I am hesitate to actually use coco's implementation.  This `check_coco.py` calculates the iou between coco algorithm and PIL for the same polygon. Most of them are fall into `[0.9-1.0]` range, but still have some in `[0.7-0.8]` range. 

But since we store the polygon directly in the data, so it is lossless, and the end user can just use `pycocotools` to render this polygon again if desired.

python/rikai/spark/types/geometry.py
eddyxu: Done

python/rikai/types/geometry.py
eddyxu: ok, changed this to `raise ValueError`

python/rikai/types/geometry.py
eddyxu: A mistake. Fixed.

python/rikai/types/geometry.py
eddyxu: Ok, fixed.

contrib/coco/scripts/make_dataset.py
eddyxu: So the purpose here is to make the output has 10 RLE and 10 polygons images.

changhiskhan: right but if you get to 10 RLE first then you won't have 10 polygon images

contrib/coco/tests/test_convert.py
changhiskhan: got it

python/rikai/types/geometry.py
changhiskhan: yeah, I understand the intended behavior, it's just the naming.
Calling it "RAW" implies it's lower-level than the other types. Off the top of my head I can suggestion something like calling the class `Segmentation` and keep the `MASK` constant name. OR keep the `Mask` class name and change the constant to `Mask.Type.DENSE` or something?

Is there community convention on this already that makes sense? If there is we can just follow it?

eddyxu: This is up to discuss. So a "Segmentation" is similar to a "Detection" IMO, where it represents the output of a "segmentation task".  I am not certain to name it Mask.TYPE.MASK or .DENSE is better.

Let me run some reference with other ppl then.

contrib/coco/scripts/make_dataset.py
eddyxu: Make sense. Fixed in the code actually.

python/rikai/types/geometry.py
eddyxu: Talked offline, lets just not support the raw / dense mask type for now.  RLE is loss less and statistically much smaller than the mask anyway. Lets always store data in RLE and polygon forms as the same as coco.

changhiskhan (approved):
LGTM provided you just remove the Mask.Types.MASK.

@darcy-shen you should take a look at this too since it's a fairly significant change.

PR #383, opened 2021-12-11
Read image as embeded

eddyxu:
And provide a conversion from external Image to embedded image

```
img = Image("s3://foo/bar.png")
img.to_embedded()
```


changhiskhan (approved):
LGTM thanks

PR #400, opened 2021-12-22
Composable annotation visualization 

eddyxu:
With the success of `Displayable` image and such types, this PR proposes a composable visualization interface, to make visualize annotations in video, image, sensor data easier in notebook and other environment.

In summary, this PR introduces three orthogonal components:

- `Drawable`, an interface to indicate that this components can be visualized. More concretely, it can be an image, an label, bounding box, segmentations and etc.
- `Draw`,  a container that contains an array of `Drawable`s. As a displayable in Jupyter notebook, Draw will render all drawbles in one single frame (i.e., one image or one lidar scan).
- `Render`, the logic to render the Drawable stack on each other, i.e., rendering bounding box and segmentation over each image. More concretely, a Render can be PIL render, SVG render or even render the `Draw` into json. 

The designed interface:s

```python
class Draw(Displayable, ToDict):
    def draw(self, drawable: Drawable) -> Draw:
         """Can chain multiple drawable together"""
         pass
   
    def __add__(self, other) -> Draw:
         return self.draw(other)

    def display(self): # Can be visualized in notebook directly
       pass

    def to_dict(self):  # Can be used to integrate with other visualization libraries
        pass

class Drawable(...):
    def render(self, render: Render, **kwargs):
         pass

class Render(...):
     def text(...): pass
     def rectangle(...): pass
     def polygon(...): pass

class PILRender(Render): pass
class SVGRender(Render): pass
class JSONRender(Render): pass
```

Similar to Apache Beam / numpy , syntax sugar will be provided to make the rendering annotations much easier, for example, you can use (`+`) to chain images with all 2d annotations (i.e., `box2d`, `mask`, `label`):

```python

img = Image(uri="s3://....")
box = Box2d(xmin=10, ymin=20, xmax=30, ymax=40)
# Render image with 3 bounding boxes
img + box1 + box2 + box3
# Or
img.draw(box1).draw(box2).draw(box3)
```

Moreover, a special `Drawable` -- `Style` is designed to make visualization customizable and composable:

```python
class Style(Drawable):
    def __call__(self, inner: Drawable): pass
    def render(self, render, **kwags): pass

cat_style = rikai.viz.Style(color="yellow",  width=4)
# Render with style
img + cat_style(cat_box) + human_style(human_box) ...
```

We will also override `@` to make style easy to use:

```python
img + cat_box@{color="yellow", width=4}
```

We can put arbitrary logic in `Style` as well:

```python

class LabelStyleMap:
   def __init__(self, color_map):
       self.color_map = color_map

   def render(self, render, **kwargs):
       self.inner.render(render, color=self.color_map[self.inner.label_id])
```
Screen Shot 2021-12-21 at 8 25 31 PM


python/rikai/mixin.py
```diff
@@ -63,6 +63,26 @@ def display(self, **kwargs) -> "IPython.display.DisplayObject":
...
 
+class Drawable(ABC):
+    """Mixin for a class that is drawable"""
+
+    @abstractmethod
+    def render(self, render: "rikai.viz.Render", **kwargs) -> None:
+        """Render the object using render."""
+
+    def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
+        from rikai.viz import Style
```
changhiskhan, line 74: this is to prevent circular import?

python/rikai/types/vision.py
```diff
@@ -209,6 +213,15 @@ def _repr_png_(self):
     def __eq__(self, other) -> bool:
         return isinstance(other, Image) and super().__eq__(other)
 
+    def __or__(self, other: Drawable) -> Draw:
+        """Override ``&`` operator to chain images with
```
changhiskhan, line 217: |  operator right?

python/rikai/types/vision.py
```diff
@@ -209,6 +213,15 @@ def _repr_png_(self):
...
+        visualization components.
+        """
+        return self.draw(other)
+
+    def render(self, render: "rikai.viz.Render", **kwargs) -> None:
```
changhiskhan, line 222: Should Image really be a Drawable?

python/rikai/viz.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+
+    def __init__(self, **kwarg):
+        self.kwargs = kwarg
+        self.inner = None  # type: Optional[Drawable]
+
+    def __repr__(self):
+        return f"style({self.kwargs})"
+
+    def __call__(self, inner: Drawable) -> Drawable:
```
changhiskhan, line 49: if we have say a colormap style, can it be applied to a list of Drawables? Or should we say you have to do [style(x) for x in drawables]?

python/rikai/viz.py
```diff
@@ -0,0 +1,150 @@
...
+        # Make a copy of Style so the same style can be applied
+        # to multiple drawables
+        s = Style(**self.kwargs)
+        s.inner = inner
+        return s
+
+    def render(self, render: Render, **kwargs):
```
changhiskhan, line 56: when would the user need to supply kwargs in the constructor and also in the render call?

python/rikai/viz.py
```diff
@@ -0,0 +1,150 @@
...
+
+class Draw(Displayable, ABC):
+    """Draw is a container that contain the elements for visualized lazily."""
+
+    def __init__(self):
+        self.layers = []
+
+    def __repr__(self):
+        first_layer = self.layers[0] if self.layers else "N/A"
+        return f"Draw({first_layer})"
```
changhiskhan, line 71: why not do all of the layers?

python/rikai/viz.py
```diff
@@ -0,0 +1,150 @@
...
+        if not isinstance(layer, Drawable):
+            raise ValueError(f"{layer} must be a Drawable")
+        self.layers.append(layer)
+        return self
+
+    def __or__(self, other: Drawable) -> Draw:
+        return self.draw(other)
+
+
+class Render(ABC):
```
changhiskhan, line 83: The naming is somewhat confusing since Draw and Render are close in meaning.
Maybe call this `Renderer` instead of Render?

changhiskhan (commented):
Can you give an example of say how we'd draw boxes with their labels styled by a color map by label?

python/rikai/mixin.py
eddyxu: yes

eddyxu:
> Can you give an example of say how we'd draw boxes with their labels styled by a color map by label?

So it requires additional implementation, some pseudo code  along the line:

```python

from rikai.viz import Style

class StyleMap(Style):
    def __init__(self, style_map: dict[str, Style]):
         super().__init__()
         self.style_map = style_map

    def __call__(self, inner: LabelledDrawable):
         super().__call__(inner)

    def render(self, render, **kwargs):
          style = self.style_map[self.inner.label_id](self.inner)
          return style.render(render, **kwargs)


style_map = StyleMap({ label: Style(color=color) for label, color in zip(labels, color_plates) })
img | style_map(box1) | style_map(box2)
```

python/rikai/viz.py
eddyxu: i should make this a private/protected methods tho. This method is not user-facing interface.

python/rikai/viz.py
eddyxu: sgtm

python/rikai/viz.py
eddyxu: Answered below

python/rikai/types/vision.py
eddyxu: Good question. lets not make `Image` a `Drawable` for now.

python/rikai/types/vision.py
eddyxu: Fixed.

python/rikai/viz.py
eddyxu: Renamed this method to `_render(..)`

python/rikai/viz.py
eddyxu: Most of the boxes and masks (segments) do not have much readable meaning via `__repr__`

changhiskhan:
> > Can you give an example of say how we'd draw boxes with their labels styled by a color map by label?
> 
> So it requires additional implementation, some pseudo code along the line:
> 
> ```python
> from rikai.viz import Style
> 
> class StyleMap(Style):
>     def __init__(self, style_map: dict[str, Style]):
>          super().__init__()
>          self.style_map = style_map
> 
>     def __call__(self, inner: LabelledDrawable):
>          super().__call__(inner)
> 
>     def render(self, render, **kwargs):
>           cur_style = self.style_map[self.inner.label_id](self.inner)
>           return cur_style.render(render, **kwargs)
> 
> 
> style_map = StyleMap({ label: Style(color=color) for label, color in zip(labels, color_palatte) })
> img | style_map(box1) | style_map(box2)
> ```

In the Coco example you don't have a definite number of boxes but instead you have a list of boxes. In this scenario, would it work like `img | list_of_boxes` ? If so how would you apply the style_map? `img | style_map(list_of_boxes)` or `img | [style_map(x) for x in list_of_boxes]` ?

changhiskhan (approved):
lgtm. This would be a lot more useful for the coco dataset if it was easier to handle:
1. Automatically placing the labels
2. A list of boxes rather than just individual ones
3. applying stylemap to the list altogether.

eddyxu:
For 

> A list of boxes rather than just individual ones
> applying stylemap to the list altogether.

Yes, we can just make `Draw.draw()` interface to take a list of `Drawable` that could work.

like

```python
class Draw:
    def draw(self, drawable: Union[Drawble, Iterator[Drawable]]) -> Draw:
          pass

Similar to Style

class Style:
    def __call__(self, inner: Union[Drawable, Iterator[Drawable]]) -> Drawable:
          pass
```

Automating placing labels is more challenging tho. :) the main concern at the moment is , if we pack label with annotations (i.e., bbox / segmentation), what is the impact of accessing it in Spark SQL, as well as the structural information of the semantic. I think we could come up something in mid term tho.

PR #404, opened 2021-12-22
Do not use height/width when construct polygon

eddyxu:
Height / width are irrelevant to polygon, and datasets like coco do not have height/width with the segmentations.

### DIFF
--- a/python/rikai/types/geometry.py
+++ b/python/rikai/types/geometry.py
@@ -514,7 +514,9 @@ class Mask(ToNumpy, ToDict):
         width: Optional[int] = None,
         mask_type: Mask.Type = Type.POLYGON,
     ):
-        if mask_type != Mask.Type.POLYGON and (height is None or width is None):
+        if mask_type != Mask.Type.POLYGON and (
+            height is None or width is None
+        ):
             raise ValueError(
                 "Must provide height and width for RLE or Polygon type"
             )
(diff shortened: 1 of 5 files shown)


python/rikai/types/geometry.py
```diff
@@ -514,7 +514,9 @@ def __init__(
         width: Optional[int] = None,
         mask_type: Mask.Type = Type.POLYGON,
     ):
-        if height is None or width is None:
+        if mask_type != Mask.Type.POLYGON and (
+            height is None or width is None
+        ):
             raise ValueError(
                 "Must provide height and width for RLE or Polygon type"
```
changhiskhan, line 521: remove "or Polygon type"?

python/rikai/types/geometry.py
```diff
@@ -572,9 +574,7 @@ def from_polygon(data: list[list[float]], height: int, width: int) -> Mask:
         width: int
```
changhiskhan, line 574: remove the docs?

changhiskhan (approved):
docstrings

python/rikai/types/geometry.py
eddyxu: fixed

python/rikai/types/geometry.py
eddyxu: fixed

PR #405, opened 2021-12-23
Draw and Style to support list of Drawables

eddyxu:
Follow up to #400

### DIFF
--- a/python/rikai/viz.py
+++ b/python/rikai/viz.py
@@ -59,7 +59,9 @@ class Style(Drawable):
 
     def _render(self, render: Renderer, **kwargs):
         if self.inner is None:
-            raise ValueError("This style object has not attack to a Drawable yet")
+            raise ValueError(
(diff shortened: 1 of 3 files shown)


python/rikai/viz.py
```diff
@@ -43,23 +43,26 @@ class Style(Drawable):
...
         # Make a copy of Style so the same style can be applied
         # to multiple drawables
         s = Style(**self.kwargs)
+        if isinstance(inner, Drawable):
+            inner = [inner]
         s.inner = inner
         return s
 
     def _render(self, render: Renderer, **kwargs):
         assert self.inner is not None
```
changhiskhan, line 61: forgot this in the last PR but asserts are removed in "production"

changhiskhan (approved):

python/rikai/viz.py
eddyxu: ok, will change this to rase ValueError then

PR #407, opened 2021-12-24
set model tag

Renkai:
Attach the tags of a run to the related model and model version.

* It seems the mlflow API changes aggressively(I guess that's because they have too many minor bugs), so we have to make the mlflow server and client version exactly the same to make the test pass. Currently, I use `1.20.2`, please comment if it's not the right version.
 https://github.com/eto-ai/rikai/runs/4625778280?check_suite_focus=true


python/tests/spark/sql/codegen/test_mlflow_logger.py
```diff
@@ -40,3 +42,120 @@ def test_none_value_tags(tmp_path: Path, resnet_model_uri):
         assert tags[rikai.mlflow.CONF_MLFLOW_OUTPUT_SCHEMA] == "output_schema"
         assert rikai.mlflow.CONF_MLFLOW_PRE_PROCESSING not in tags
         assert rikai.mlflow.CONF_MLFLOW_POST_PROCESSING not in tags
+
+
+def test_model_and_version_tags(tmp_path: Path, resnet_model_uri):
```
changhiskhan, line 47: this is one huge test case, can we break this into smaller pieces?

changhiskhan: for example, you can factor out the model assertions into a function and call it repeatedly.

python/tests/spark/sql/codegen/test_mlflow_logger.py
```diff
@@ -40,3 +42,120 @@ def test_none_value_tags(tmp_path: Path, resnet_model_uri):
...
+    tmp_path.mkdir(parents=True, exist_ok=True)
+    tracking_uri = "sqlite:///" + str(tmp_path / "tracking.db")
+    mlflow.set_tracking_uri(tracking_uri)
+    model_name = "model_for_tags"
+    c = MlflowClient()
+    c.create_registered_model(model_name)
+    with mlflow.start_run():
```
changhiskhan, line 54: this block that logs a model can probably be reused as well.

python/setup.py
```diff
@@ -31,7 +31,7 @@
 docs = ["sphinx"]
 video = ["ffmpeg-python", "scenedetect"]
 youtube = ["pafy", "youtube_dl"]
-mlflow = ["mlflow>=1.15"]
+mlflow = ["mlflow==1.20.2"]
```
changhiskhan, line 34: is it safe to pin it to the minor version at least?

python/rikai/spark/sql/codegen/mlflow_logger.py
```diff
@@ -131,6 +132,30 @@ def log_model(
...
+            for v in all_versions:
+                if v.run_id == run_id:
+                    current_version = v
+                    break
+            if current_version is None:
+                raise ValueError(
+                    "No model version found matching runid: {}".format(run_id)
+                )
+            for key, value in tags.items():
+                c.set_registered_model_tag(registered_model_name, key, value)
```
changhiskhan, line 155: if the previous version logged some tags that's removed in this version, is there no way to detect/remove?

changhiskhan:
oh btw @Renkai there was also a TODO item from the persistent mlflow catalog PR that was waiting for this. Can you take a look and see if it makes sense to address it together in this PR?

python/rikai/spark/sql/codegen/mlflow_logger.py
Renkai: It's possible, but we need to define the `previous version` first. Is it the previous version of current model name? It's possible the previous version is generated by other unrelated runs.
And then, maybe we should define a rikai related tags scope like what tags are managed by rikai to prevent other tags accidentally deleted by rikai

python/setup.py
darcy-shen: not safe, as mentioned in the pull request body

Renkai:
> oh btw @Renkai there was also a TODO item from the persistent mlflow catalog PR that was waiting for this. Can you take a look and see if it makes sense to address it together in this PR?

Make sense, updated

python/setup.py
Renkai: From the docs https://mlflow.org/docs/latest/rest-api.html#get-latest-modelversions
and my local curl tests. The mlflow server doesn't obey the docs (docs say it should be POST, but it's actually GET), so we must choose the client version that also doesn't obey the docs. We can't tell if a client version works as we expected before a test.
```
(rikai) ➜  python git:(renka-set-model-tag) ✗ curl -X GET http://localhost:5000/api/2.0/mlflow/registered-models/get-latest-versions -d '{"name":"jb", "stages": []}'
{"error_code": "RESOURCE_DOES_NOT_EXIST", "message": "Registered Model with name=jb not found"}%
(rikai) ➜  python git:(renka-set-model-tag) ✗ curl -X POST http://localhost:5000/api/2.0/mlflow/registered-models/get-latest-versions -d '{"name":"jb", "stages": []}'
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>405 Method Not Allowed</title>
<h1>Method Not Allowed</h1>
<p>The method is not allowed for the requested URL.</p>
```

changhiskhan:
@Renkai looks like tests failed, can you take a look?

Renkai:
> @Renkai looks like tests failed, can you take a look?

Investigating, it runs fine on my computer, weird😂.

python/rikai/spark/sql/codegen/mlflow_logger.py
changhiskhan: Please create a GH issue to deal with the scoping (e.g., use a prefix, turn the rikai tags into a single dict tag). Please describe the problem fully and link to this so later people know what it's about

python/setup.py
changhiskhan: lol mlflow ftl

Renkai:
added a new issue https://github.com/eto-ai/rikai/issues/420

changhiskhan (approved):
lgtm

Issue #410, opened 2021-12-26
Type conversion bug?


changhiskhan:
@eddyxu @darcy-shen not sure if i'm reading the code correctly but:

https://github.com/eto-ai/rikai/blob/24710a47da651e7ba650971debd93562fc7f7b96/python/rikai/parquet/dataset.py#L153

If a column is a plain struct that has a UDT field, i don't think it will be converted right?

PR #411, opened 2021-12-26
[PARQUET] Fix a bug converting udt subfield of struct field

changhiskhan:
And also add a `to_pandas` method to rikai.parquet.dataset.Dataset

### DIFF
--- a/python/rikai/parquet/dataset.py
+++ b/python/rikai/parquet/dataset.py
@@ -16,12 +16,15 @@
 """
 
 # Standard Library
+from functools import partial
 import importlib
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
 # Third Party
-import pyarrow.parquet as pg
+import pandas as pd
+from pyarrow import fs
+import pyarrow.parquet as pq
 from pyarrow.fs import FileSystem
 from pyspark.ml.linalg import Matrix, Vector
 from pyspark.sql import Row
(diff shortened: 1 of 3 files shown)


changhiskhan:
closes #410

python/rikai/parquet/dataset.py
```diff
@@ -228,3 +225,50 @@ def __iter__(self):
...
+        if not isinstance(field_type, dict):
+            return col
+        elif field_type["type"] == "udt":
+            udt = self._find_udt(field_type["pyClass"])
+            return col.apply(partial(_convert_udt_value, udt=udt))
+        elif field_type["type"] == "struct":
+            return col.apply(lambda d: self._convert(d, field_type))
+        elif field_type["type"] == "array":
+
+            def convert_array(arr):
```
eddyxu, line 256: shall we move this to be a standalone function globally?

python/rikai/parquet/dataset.py
```diff
@@ -228,3 +225,50 @@ def __iter__(self):
             yield self._convert(
                 shuffler.pop().to_dict(), self.spark_row_metadata
             )
+
+    def to_pandas(self):
```
eddyxu, line 229: is this user-facing API?  lets add a comment to mention that this method loads all data into memory?

changhiskhan: I'll add some docstrings.
I was thinking I can add a `limit` parameter that's just pq.ParquetDataset.head (https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Dataset.html#pyarrow.dataset.Dataset.head) if you only want a preview

python/rikai/parquet/dataset.py
changhiskhan: yeah, good idea this is usable for the self._convert as well

changhiskhan:
@eddyxu take another look?

eddyxu (approved):
LGTM. Thanks @changhiskhan

PR #419, opened 2021-12-27
Remove circe by json4s

da-tubi:


src/main/scala/ai/eto/rikai/sql/model/Model.scala
```diff
@@ -17,11 +17,11 @@
 package ai.eto.rikai.sql.model
 
 import ai.eto.rikai.sql.spark.SparkRunnable
-import io.circe.syntax._
-import io.circe.generic.encoding.DerivedAsObjectEncoder._
 import org.apache.spark.sql.catalyst.FunctionIdentifier
 import org.apache.spark.sql.catalyst.analysis.UnresolvedFunction
 import org.apache.spark.sql.catalyst.expressions.Expression
+import org.json4s.DefaultFormats
```
eddyxu, line 23: should we explicitly declare json4s in `build.sbt` instead of relying on transient dependency?

da-tubi: No. Just rely on transient dependencies. json4s (jackson binding) is heavily used in Apache Spark.

If we explicitly declare json4s and mark it as runtime, we need to align the version with Apache Spark.

eddyxu: So what if Spark decides to upgrade to an incompatible version of `json4s` upstream and breaks `rikai` silently then? Will Spark owns the API-stability of its own transient dependency in this case? 

Or what if some security issues similar to log4j happen to `json4s`, but we dont have the chance to upgrade Spark in a short period of time because other incompatibility in Spark (i.e., similar to Spark 3.2 to Spark 3.1) or other libraries for a lot of reasons, like internal API changes, scala 2.12 to 2.13 or etc. 

In general, i don't think it is a good practice that we can not fully control the version of libraries we used explicitly.

darcy-shen: In this specific case, we are using json4s only to convert a Map[String, String] to json. We could avoid security in rikai easily.

> Or what if some security issues similar to log4j happen to json4s

If we do not use json4s, this problem still exists.
If it happens, we could switch to another light weighted json library. Or concat the Map into json by hand.

> In general, i don't think it is a good practice that we can not fully control the version of libraries we used explicitly.
Here, we are only using the json library to convert a Map to a string. The library is not heavily used, I think in this case, it is ok to use the transient dependency.

da-tubi: In this specific case, we are using json4s only to convert a Map[String, String] to json. We could avoid security in rikai easily.

> Or what if some security issues similar to log4j happen to json4s

If we do not use json4s, this problem still exists.
If it happens, we could switch to another light weighted json library. Or concat the Map into json by hand.

> In general, i don't think it is a good practice that we can not fully control the version of libraries we used explicitly.

Here, we are only using the json library to convert a Map to a string. The library is not heavily used, I think in this case, it is ok to use the transient dependency.

da-tubi: > Or what if some security issues similar to log4j happen to json4s

If we use another json library, actually we are doubling the security risk.

da-tubi: To clarify, `json4s` is a wrapper here. The underlying library to serialize Map[String, String] is jackson.

eddyxu: My main concern is not particularly to `json4s` per se. It is more about long term maintainability or the practice should be applied to this project. Do we have a clear line of what should be transient dependency and what should not? Should we not include any library that one of our main dependencies has (for example, just use `antlr` from Spark as well)? 

For an early stage project like `rikai`, once an essential library is settled, it is really hard to say that it will only be used for `Map[string, string] to json conversion` in long short, especially for such an essential function (json).

Without explicitly to declare a library, WHEN and HOW a library was upgraded is completely in the dark from rikai's devs, which either be caught on CI when we upgrade the dependency that introduces them (in this case, Spark) if we are lucky and have really decent test coverage. Or it just blows up in the production that is visible to potential users.  Do we want to pay the cost every time Spark or some other open source community spent a month to upgrade `json4s`/`antlr`/ `snappy` and etc? 

Using Spark internal API for SQL parser is an (unfortunate) cost that we have to pay to offer Spark SQL experience. But I dont see that not directly using a public available library for stability is such a risk we should pay. 

Additionally, for vendors like AWS EMR / Databricks / Cloudera, it is not unusual to ship / re-compile certain components differently from the open source projects for various reasons.  Similarly, RedHat and Ubuntu will backport security fixes / major bug fixes into old LTS kernel too (which means the LTS kernel has been different to the open source point releases once the LTS kernel branch is cut).  Explicitly declaring the libraries that Rikai directly uses, gives the dev full control of WHEN and HOW to upgrade.  

In this example, if any reason that we need to upgrade a library in response of certain events, we can upgrade it faster (For example, in some hypothetic future, we were stuck on Spark 3 but the security fix only happens in Spark 4) , and rely on the vendors like EMR / DataBricks to patch Spark runtime while maintain the same spark version (i.e., still Spark 3.1.2 but uses a security fixed json4s which might or might not compatible)

da-tubi: To avoid depending on Spark, we could shade it when there are conflicts or it is necessary.

When we want full control on WHEN and HOW to upgrade, shading is the final solution.

da-tubi: Apache Flink did a good job on shading dependencies: https://github.com/apache/flink

changhiskhan:
let's revisit dependency convention later

PR #431, opened 2021-12-29
udf: array_box2d

da-tubi:
Demo usage of `array_box2d`: https://github.com/da-tubi/rikai-example/blob/54ce6f71e12be369b3bbd197d34f034e2d01d141/elephants_dream.ipynb

### DIFF
--- a/python/rikai/spark/functions/__init__.py
+++ b/python/rikai/spark/functions/__init__.py
@@ -31,6 +31,7 @@ def init(spark: SparkSession):
     all_geo_udfs = [
         "area",
         "box2d",
+        "array_box2d",
         "box2d_from_center",
         "box2d_from_top_left",
     ]
--- a/python/rikai/spark/functions/geometry.py
+++ b/python/rikai/spark/functions/geometry.py
@@ -14,22 +14,39 @@
 
 """Geometry related PySpark UDF"""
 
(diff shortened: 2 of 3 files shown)


changhiskhan:
Looking at the notebook you're using this udf to convert the ML_PREDICT output into Rikai types. Why not do that in the `post_processing`?

The major question here is how do we provide a generalized mechanism to convert to Rikai types. Do we want to have a `array_<rikai type>` udf for every Rikai type? Typically when/how are we expecting our users to do the type conversion themselves?

eddyxu:
Can we use standard spark sql function instead?

https://spark.apache.org/docs/latest/api/sql/index.html#transform

```sql
SELECT transform(boxes, b -> box2d(b)) FROM ...
```

da-tubi:
Neither `ML_PREDICT` nor `TRANSFORM` supports UDT as the return data type.

da-tubi:
> Can we use standard spark sql function instead?

No. The reason is listed above.

eddyxu:
Shall we fix `ML_PREDICT` **AND** `TRANSFORM` then?

eddyxu:
This is the fix for `ML_PREDICT` 

https://github.com/eto-ai/rikai/pull/448

PR #434, opened 2021-12-30
Tune image display on Databricks notebook again

### LINKED ISSUE #357: Databricks Notebook support for Image.display

da-tubi:
Fix #357 

+ [x] Tested on Databricks Notebook
+ [x] Tested on Jupyter Notebook

## Test Cases for Databricks Notebook
image

image

image

## Test Cases for Jupyter Notebook

image

image

``` python
from rikai.types.vision import Image

Image("s3://bucket/darcy/image.png")
```
This case on local jupyter notebook does not work.
It is another issue that we need to fix later at https://github.com/eto-ai/rikai/issues/435.

### DIFF
--- a/python/tests/types/test_vision.py
+++ b/python/tests/types/test_vision.py
@@ -35,20 +35,6 @@ def test_image() -> PILImage:
     return PILImage.fromarray(rescaled)
 
 
-def test_show_embedded_png(tmp_path, test_image):
-    uri = tmp_path / "test.png"
-    test_image.save(uri)
-    result = Image(uri)._repr_png_()
-    assert result is None
-
-
-def test_show_embedded_jpeg(tmp_path, test_image):
-    uri = tmp_path / "test.jpg"
-    test_image.save(uri)
-    result = Image(uri)._repr_jpeg_()
-    assert result is None
-
-
 def test_convert_to_embedded_image(tmp_path, test_image: PILImage):
     uri = tmp_path / "test.jpg"
     test_image.save(uri)
--- a/python/tests/types/test_vision.py
+++ b/python/tests/types/test_vision.py
@@ -39,14 +39,14 @@ def test_show_embedded_png(tmp_path, test_image):
     uri = tmp_path / "test.png"
     test_image.save(uri)
     result = Image(uri)._repr_png_()
-    assert result == None
+    assert result is None
 
 
 def test_show_embedded_jpeg(tmp_path, test_image):
     uri = tmp_path / "test.jpg"
     test_image.save(uri)
     result = Image(uri)._repr_jpeg_()
-    assert result == None
+    assert result is None
 
 
 def test_convert_to_embedded_image(tmp_path, test_image: PILImage):
--- a/python/tests/types/test_vision.py
+++ b/python/tests/types/test_vision.py
@@ -23,6 +23,7 @@ import pytest
 from PIL import Image as PILImage
 from PIL import ImageDraw
 
+from rikai.io import open_uri
 from rikai.types.geometry import Box2d
 from rikai.types.vision import Image
 from rikai.viz import Style, Text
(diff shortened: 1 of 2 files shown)


python/rikai/viz.py
```diff
@@ -21,7 +21,7 @@
 from PIL import Image as PILImage
 from PIL import ImageDraw
 
-from rikai.conf import get_option, CONF_RIKAI_VIZ_COLOR
+from rikai.conf import CONF_RIKAI_VIZ_COLOR, get_option
```
da-tubi, line 24: by `make fix`

python/tests/types/test_vision.py
```diff
@@ -39,26 +39,14 @@ def test_show_embedded_png(tmp_path, test_image):
...
     test_image.save(uri)
     result = Image(uri)._repr_jpeg_()
-    with open(uri, "rb") as fh:
-        expected = b2a_base64(fh.read()).decode("ascii")
-        assert result == expected
-
-        fh.seek(0)
-        embedded_image = Image(fh)
-        assert result == embedded_image._repr_jpeg_()
+    assert result is None
```
eddyxu, line 49: why `_repr_jpeg_` and `_repr_png_` returns None?
Are they expected? if so,  why do we still keep these two tests?

changhiskhan: let's remove these two tests and check which method jupyter/databricks is actually using and test those? I don't want to have to rely on visual inspection every time to ensure this isn't broken

da-tubi: They are not expected. I think it is a Databricks notebook bug.

da-tubi: > check which method jupyter/databricks is actually using and test those

I have no idea on how to check and how to test.

changhiskhan: At least for Jupyter you should be able to see which _repr_* method is being called by IPython display. That's how I came up with these two tests. I think databricks we can't do anything about since the notebook is closed source

da-tubi: For jupyter, I think a valid `IPython.display.Image` return object is sufficient for image displaying.

Personally, I do not know much about how jupyter is implemented. I do not want to spend too much time figuring out how IPython render an Image from `IPython.display.Image`.

I've removed the tests. I think this pr is good enough.

changhiskhan:
@da-tubi please add tests for jupyter. these are the most direct form of "user experience" features so it's important we test these for jupyter notebooks at least. ping me on slack/wechat if you want to discuss how to do it.

da-tubi:
OK, I've mark it pull request as draft. Let me learn about jupyter first.

da-tubi:
https://github.com/ipython/ipython/blob/f162b26d9158838e1d8051f790d877e85c7737b1/IPython/core/display.py#L991

I think I should test `_repr_html_ `.

da-tubi:
The main idea of the first approach is that you have to implement special
    display methods when you define your class, one for each representation you
    want to use. Here is a list of the names of the special methods and the
    values they must return:
      - `_repr_html_`: return raw HTML as a string, or a tuple (see below).
      - `_repr_json_`: return a JSONable dict, or a tuple (see below).
      - `_repr_jpeg_`: return raw JPEG data, or a tuple (see below).
      - `_repr_png_`: return raw PNG data, or a tuple (see below).
      - `_repr_svg_`: return raw SVG data as a string, or a tuple (see below).
      - `_repr_latex_`: return LaTeX commands in a string surrounded by "$",
                        or a tuple (see below).
      - `_repr_mimebundle_`: return a full mimebundle containing the mapping
                             from all mimetypes to data.
                             Use this for any mime-type not listed above.

da-tubi:
https://github.com/ipython/ipython/blob/f162b26d9158838e1d8051f790d877e85c7737b1/IPython/core/formatters.py#L925-L976

`_repr_mimebundle_` is the function which is actually used.

da-tubi:
@changhiskhan @eddyxu  ready to be reviewed

changhiskhan (approved):

PR #440, opened 2022-01-01
Eliminate the needs to start a Pyspark session to use SQL ML

eddyxu:
To make Rikai more suitable for a service, this PR  re-architecture the way that Spark SQL engine manages ML model code from Python.

We can start a Spark SQL with `ML_Predict` without starting `pyspark`, as long as we have the configuration in place (i.e., in `spark-defaults.conf`) and set propriety Python versions in driver and worker nodes.  

```
spark.sql.extensions ai.eto.rikai.sql.spark.RikaiSparkSessionExtensions
```

Spark can pick up the right udf to run python models, like pytorch, tf and sklearn. It could open more use cases / deployment modes for Rikai, i.e., can use Rikai with Livy, or Spark thrift server then


eddyxu:
Update: Only works for FileSystemRegistry model spec, will follow up with MLflowRegistry and others later in the following days.

src/main/scala/org/apache/spark/sql/rikai/model/Resolver.scala
```diff
@@ -0,0 +1,81 @@
...
+                 |    fobj.write(pickle.dumps((func, dataType)))
+                 |with open("${dataTypePath}", "w") as fobj:
+                 |    fobj.write(dataType.json())
+                 |""".stripMargin)
+      val cmd = Files.readAllBytes(path)
+      val dataTypeJson = Files.readString(dataTypePath)
+      val returnType = DataType.fromJson(dataTypeJson)
+      val udf =
+        UserDefinedPythonFunction(
+          "test_sum",
```
eddyxu, line 59: ignore these naming details for now. The current state proves that the mechanism works

eddyxu:
Ran `sbt "testOnly **.MLPredictTest"`

```
+--------------------+
|                   s|
+--------------------+
|{[[3.70917, 230.8...|
+--------------------+
root
 |-- s: struct (nullable = true)
 |    |-- boxes: array (nullable = true)
 |    |    |-- element: array (containsNull = true)
 |    |    |    |-- element: float (containsNull = true)
 |    |-- scores: array (nullable = true)
 |    |    |-- element: float (containsNull = true)
 |    |-- label_ids: array (nullable = true)
 |    |    |-- element: integer (containsNull = true)
```

src/main/scala/org/apache/spark/sql/rikai/model/Resolver.scala
```diff
@@ -0,0 +1,89 @@
...
+                 |""".stripMargin,
+        session
+      )
+      val cmd = Files.readAllBytes(path)
+      val dataTypeJson = Files.readString(dataTypePath)
+      val returnType = DataType.fromJson(dataTypeJson)
+      val suffix = Random.alphanumeric.take(6)
+      val udfName = s"${spec.name}_${suffix}"
+      val udf =
+        UserDefinedPythonFunction(
```
changhiskhan, line 64: can we also use this mechanism to expose the other rikai pyspark udfs? (doesn't have to be done in this PR)

src/main/scala/org/apache/spark/sql/rikai/model/Resolver.scala
```diff
@@ -0,0 +1,89 @@
...
+      session: SparkSession,
+      registryClassName: String,
+      spec: ModelSpec
+  ): SparkUDFModel = {
+    val specPath = Files.createTempFile("model-spec", ".json")
+    val path = Files.createTempFile("model-code", ".cpt")
+    val dataTypePath = Files.createTempFile("model-type", ".json")
+    try {
+      Files.writeString(specPath, spec.asJson.toString)
+      Python.execute(
```
changhiskhan, line 44: what does the error look like here if say the spec is bad?

src/main/scala/org/apache/spark/sql/rikai/model/Resolver.scala
```diff
@@ -0,0 +1,89 @@
...
+            Python.pythonExec,
+            Python.pythonVer,
+            Seq.empty.asJava,
+            null
+          ),
+          returnType,
+          PythonEvalType.SQL_SCALAR_PANDAS_ITER_UDF,
+          udfDeterministic = true
+        )
+      session.udf.registerPython(udfName, udf)
```
changhiskhan, line 79: is this being done on every invocation? or just the first time when this udf is created?

changhiskhan (commented):
very exciting! Some questions for my own clarification

src/main/scala/org/apache/spark/sql/rikai/model/Resolver.scala
eddyxu: yes, so basically just use the same python interpreter to pickle function and its return `DataType`, you can register any functions (with some size limitations).

This is how pyspark registers udfs and ships them to workers.

src/main/scala/org/apache/spark/sql/rikai/model/Resolver.scala
eddyxu: It throws a `RuntimeError(captured_stderr)`

src/main/scala/org/apache/spark/sql/rikai/model/Resolver.scala
eddyxu: this is everytime a model / udf is created / registered.

python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -187,11 +183,11 @@ def tracking_client(self):
 
     @property
     def mlflow_tracking_uri(self):
-        return self._spark.conf.get(CONF_MLFLOW_TRACKING_URI)
+        return os.environ.get(CONF_MLFLOW_TRACKING_URI)
```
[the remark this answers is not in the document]
eddyxu: So the JIT codegen does not have SparkSession, it is just a pure python process to pickle python code.

python/rikai/spark/sql/codegen/base.py
```diff
@@ -201,34 +201,8 @@ def udf_from_spec(spec: ModelSpec):
...
-    Returns
-    -------
-    str
-        Spark UDF function name for the generated data.
-    """
-    udf = udf_from_spec(spec)
-    return register_udf(spark, udf, name)
+def command_from_spec(registry_class: str, row_spec: dict):
+    cls = find_class(registry_class)
+    resolver = cls()
```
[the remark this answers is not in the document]
eddyxu: lemme rename this to `registry`

src/main/scala/ai/eto/rikai/sql/spark/Python.scala
```diff
@@ -16,52 +16,47 @@
...
+import scala.sys.process
+import scala.sys.process.{Process, ProcessLogger}
 
 object Python {
-  private var python: Option[Python] = None
 
-  def register(mr: Python): Unit =
-    python = Some(mr)
+  /** Python executor */
+  lazy val pythonExec: String =
```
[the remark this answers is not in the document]
eddyxu: None that is obvious to me. 

Guess that was the name that Spark developers made something similar in their tests

https://github.com/apache/spark/blob/11596b3b17b5e0f54e104cd49b1397c33c34719d/sql/core/src/test/scala/org/apache/spark/sql/IntegratedUDFTestUtils.scala#L197

changhiskhan (approved):
This is awesome

PR #443, opened 2022-01-02
Remove torch DataLoader for 0.1 release

eddyxu:
`rikai.torch.data.DataLoader` should be deprecated after pytorch 1.8. Given that Rikai depends on pytorch 1.8+, it is safe to remove `rikai.torch.data.DataLoader` now.


changhiskhan (approved):
Hard to review the nb changes but yay for removing code :)

PR #445, opened 2022-01-02
Tensorflow dataset

eddyxu:


python/setup.py
```diff
@@ -25,14 +25,26 @@
...
+tf = ["tensorflow>=2.5.0"]
 jupyter = ["matplotlib", "jupyterlab"]
 aws = ["boto3", "botocore"]
 gcp = ["gcsfs"]
 docs = ["sphinx"]
 video = ["ffmpeg-python", "scenedetect"]
 youtube = ["pafy", "youtube_dl"]
 mlflow = ["mlflow==1.20.2"]
-all = sklearn + torch + jupyter + gcp + docs + video + youtube + mlflow + aws
+all = (
```
changhiskhan, line 36: Should we discourage users from installing all? Feels like it's really heavy to install both frameworks and also both GCP and aws support etc

python/tests/tf/test_tf_data.py
```diff
@@ -0,0 +1,47 @@
...
+        image = Image.from_array(image_data)
+
+        data.append(Row(id=i, image=image))
+    df = spark.createDataFrame(data)
+    df.write.format("rikai").save(str(dataset_dir))
+
+    dataset = from_rikai(
+        dataset_dir,
+        output_signature=(
+            tf.TensorSpec(shape=(), dtype=tf.uint8),
```
changhiskhan, line 37: Ok I see what you mean about the inconvenience

python/rikai/tf/data.py
```diff
@@ -0,0 +1,54 @@
...
+    output_signature: Tuple,
+    columns: Optional[List[str]] = None,
+):
+    """Build a Tensorflow (tf) Dataset from rikai dataset.
+
+    Parameters
+    ----------
+    query : str
+        A dataset URI
+    output_signature : tuple
```
changhiskhan, line 39: Is the plan that we will try to infer the schema in the future?

changhiskhan (commented):
A little concerned that the dataset api is different between tf vs pytorch

eddyxu:
> A little concerned that the dataset api is different between tf vs pytorch

I tried to model this api from [`from_generator`](https://www.tensorflow.org/api_docs/python/tf/data/Dataset#from_generator)  API, so that it gives a native`tensorflow` UX , instead of native `Rikai` UX. Prob we can model it after `tf.data.TFRecordDataset` too i guess.

Similarly, we've done `torch` Dataset after `torch.util.data`.

python/setup.py
eddyxu: Yea, it is good call. I am open to it. This is really way to much to install.

python/rikai/tf/data.py
eddyxu: Really wanted to do so. Tried a few days earlier last year, don't see a good elegant solution without users either especially specifying the `schema / spec / shape` either in producer side (in feature engineering) or consumer side (data loader).

The main mismatch is that I could not find a way to define parameterized user defined types in Spark, so that it is hard to persist shapes, especially for the `Numpy array` type in the dataset.  

So instead, if we wanted to persist the schema in the producer side, we can probably do something like

```
df.write.format("rikai).option("rikai.data.spec", SomeSpec.to_json()).save("s3://foo/bar")
```

changhiskhan (approved):
I think the import optimization and schema metadata improvement can be done separately

PR #446, opened 2022-01-03
Use subprocess for mlflow

Renkai:
This pr fix https://github.com/eto-ai/rikai/compare/use_subprocess_for_mlflow?expand=1
Should save time for about 1 min in every test workflow.

### DIFF
--- a/python/tests/spark/sql/codegen/conftest.py
+++ b/python/tests/spark/sql/codegen/conftest.py
@@ -16,11 +16,11 @@ from __future__ import annotations
 
 import datetime
 import os
+import subprocess
 
 import mlflow
 import pytest
 import torch
-import torchvision
 from mlflow.tracking import MlflowClient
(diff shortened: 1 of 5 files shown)


python/tests/spark/sql/codegen/conftest.py
```diff
@@ -16,22 +16,47 @@
...
+            "--workers",
+            "1",
+            "--backend-store-uri",
+            db_url,
+            "--default-artifact-root",
+            dir_url,
+        ]
+    )
+    yield p
+    p.kill()
```
changhiskhan, line 52: Does this need to be in a finally block?

Renkai:
Blocked by some problems, subprocess can only provide service for python tests, but scala tests also need mlflow service. So it seems docker-compose is still the best way to provide mlflow services for all the tests.😅

PR #459, opened 2022-01-05
use hosted toolcache to replace ~/.cache/pip

Renkai:
From the logs here: https://github.com/eto-ai/rikai/runs/4714645962?check_suite_focus=true#step:11:13
We can see not all pip packages will go through `~/.cache/pip`, but all pip packages will go through `/opt/hostedtoolcache/Python/`.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -53,10 +53,12 @@ jobs:
     - name: Pip install
       working-directory: python
       run: |
+        PYTHONUSERBASE=/opt/hostedtoolcache/rikai/${{ matrix.python-version }} \
         python -m pip install --target=/opt/hostedtoolcache/rikai/${{ matrix.python-version }} -e .[all,dev]
     - name: Run python tests
       working-directory: python
       run: |
+        pip install pytest
         PYTHONUSERBASE=/opt/hostedtoolcache/rikai/${{ matrix.python-version }} pytest -x -v --durations=10 --ignore=tests/parquet/internal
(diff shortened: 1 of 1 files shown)


darcy-shen:
cool, from 2m -> 5s

changhiskhan (approved):
does this cache all of the extras in setup.py as well?

Renkai:
> does this cache all of the extras in setup.py as well?

I guess so if the `setup.py` has not installed things to somewhere outside their expected path, but this patch still has some issue, if he `actions/setup-python@v2` silently upgrade its python patch version, the cache would be invalidated and the cache action can't update it. I'm finding a way to solve this issue.

Renkai:
After the last commit. I guess it's already the best pip cache solution in the github universe.

* We cache the installed package, not the downloaded package. Because `pip install` spends lots more time in installing than just downloading.
* We install the packages into the user namespace, separate from the global namespace so that the cache doesn't need to include the standard library of Python(about `1GB` per interpreter)
* The total cache size is about 5GB(sum of three Python interpreters), still far from the github limit of 10GB, I believe we can hang on to next time github extends its cache limit.
* Total time of cache and pip install reduced from about 3 minutes to 1 minute.

Renkai:
> does this cache all of the extras in setup.py as well?

It will save the cache directory when the PR workflow is finished, and we have already installed all extras in the workflow, so the answer is yes.

darcy-shen (approved):

PR #464, opened 2022-01-06
Fix ImageDraw and clean unused repr for Image

### LINKED ISSUE #462: ImageDraw broken on Jupyter Notebook
due to #434 

Here is the notebook to reproduce it: https://github.com/da-tubi/rikai-example/blob/39ab170546ca7c6f7b66787aff4a36ccb9fa01d8/elephants_dream.ipynb

Actions:
+ Fix it and test it on Jupyter Notebook and Databricks Notebook
+ Add unit tests on _repr_mimebundle_

da-tubi:
Fixes #462 

For arbitrary python object, `_repr_mimebundle_` is the magic method which is actually invoked for displaying in Jupyter Notebook.

### DIFF
--- a/python/rikai/types/vision.py
+++ b/python/rikai/types/vision.py
@@ -190,23 +190,11 @@ class Image(ToNumpy, ToPIL, Asset, Displayable, ToDict):
             return "Image(<embedded>)"
         return f"Image(uri={self.uri})"
 
-    def _repr_html_(self):
-        """Default visualizer for remote ref (or local ref under cwd)"""
-        return self.display()._repr_html_()
-
     def _repr_mimebundle_(self, include=None, exclude=None):
         """default visualizer for embedded mime bundle"""
         return self.display()._repr_mimebundle_(
             include=include, exclude=exclude
         )
-
-    def _repr_jpeg_(self):
-        """default visualizer for embedded jpeg"""
-        return self.display()._repr_jpeg_()
-
-    def _repr_png_(self):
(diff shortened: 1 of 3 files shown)


python/rikai/types/vision.py
```diff
@@ -190,24 +190,12 @@ def __repr__(self) -> str:
             return "Image(<embedded>)"
         return f"Image(uri={self.uri})"
 
-    def _repr_html_(self):
```
changhiskhan, line 193: Have you tested the full combination of (externalized, embedded) x (Image, ImageDraw ) x (databricks, Jupyter)?

python/rikai/viz.py
```diff
@@ -79,6 +79,12 @@ def __repr__(self):
         first_layer = self.layers[0] if self.layers else "N/A"
         return f"Draw({first_layer})"
 
+    def _repr_mimebundle_(self, include=None, exclude=None):
```
changhiskhan, line 82: oic for display you've converted everything into an embedded repr for databricks notebooks so we have to implement this

changhiskhan (approved):
Can you please confirm that you checked the full combo works (8 different options)?

python/rikai/types/vision.py
da-tubi: I've already tested it for Jupyter. Let me verify the other 4 cases for Databricks.

da-tubi: confirmed on Databricks now!

PR #470, opened 2022-01-07
Docs for TorchHub registery

da-tubi:


docs/sqlml.rst
```diff
@@ -107,6 +109,66 @@ the content of "s3://bucket/to/spec.yaml" can be:
...
+    .. code-block:: verbatim
+
+        torchhub:///repo_owner/repo_name[:tag_name]/model_name
+
+
+In this case, here is the corresponding Python snippets to load the model:
+
+    .. code-block:: python
+
+        model = torch.hub.load('pytorch/vision:v0.9.1', 'resnet50', pretrained=True)
```
eddyxu, line 149: Can this integration load non-pretrained model?

docs/sqlml.rst
```diff
@@ -107,6 +109,66 @@ the content of "s3://bucket/to/spec.yaml" can be:
...
+
+
+Given the `repo_owner`, `repo_name` and `model_name`, here is how default PREPROCESSOR and
+POSTPROCESSOR are generated:
+
+    .. code-block:: SQL
+
+        PREPROCESSOR 'rikai.contrib.torchhub.{repo_owner}.{repo_name}.{model_name}.pre_processing'
```
eddyxu, line 157: what if the model owner in torchhub changes the input/output for a model between two versions / tags?

docs/sqlml.rst
darcy-shen: I have not tested it. It is not planned in this IMPL for Torchhub registery.

docs/sqlml.rst
darcy-shen: let the user get their hands dirty to handle different version of models

torchhub registery is for exploring

for rikai shipped pre/post processors, there are unit tests to make sure it works with the version shipped by Rikai.

eddyxu: What is user story here? My impression of a user journey is that
 

1. User tries the torchhub model over some coco compatible dataset, to visualize coco-defined labels.
2. Train the model using users' own dataset for another purpose (i.e., https://github.com/eto-ai/rikai/issues/456) 
3. Evaluate, debug the torchhub model + user's own dataset, preferably via `ML_PREDICT`
4. Deploy to production

It looks like the `torchhub://` is designed for 1,  how can user do the rest which might be the majority of their workflow?

darcy-shen: Yes.

`torchhub://` is for new Rikai users to get themselves bootstrapped with minimum dependencies.

eddyxu (approved):
I still have some questions w.r.t how this fit into the end to end workflow tho. But not a blocker.

PR #474, opened 2022-01-08
Provide Python API to register model

### LINKED ISSUE #350: Ad-hoc pre_processor and post_processor in Jupyter/Databricks Notebook
Define a pre_processor in notebook, it won't be available in PySpark executor.

eddyxu:
Close #350

### DIFF
--- a/python/rikai/spark/model.py
+++ b/python/rikai/spark/model.py
@@ -1,0 +1,44 @@
+#  Copyright (c) 2022 Rikai Authors
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""Wraps SQL ML functionalities in Python.
+"""
+
+from pathlib import Path
+from typing import Callable, Optional, Union
+import logging
+
+from pyspark.sql import SparkSession
+from pyspark.sql.types import DataType
+
+
+def create_model(
+    name: str,
+    model_uri: Union[str, Path],
+    schema: Union[str, DataType],
+    preprocessor: Union[str, Callable] = None,
+    postprocessor: Optional[Union[str, Callable]] = None,
+    replace_if_exist: Optional[bool] = False,
+):
+    active_session = SparkSession.getActiveSession()
+    assert (
+        active_session is not None
+    ), "Must run create_model with an active SparkSession"
+    print(active_session)
+
+    logging.info("Register model=%s uri=%s schema=%s", name, model_uri, schema)
+
+
+def list_models():
+    pass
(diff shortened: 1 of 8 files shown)


python/rikai/spark/model.py
```diff
@@ -0,0 +1,116 @@
...
+    jvm = active_session._jvm
+    jvm.ai.eto.rikai.sql.spark.parser.RikaiExtSqlParser.initRegistry(
+        jvm.SparkSession.getActiveSession().get()
+    )
+    cmd = jvm.ai.eto.rikai.sql.spark.execution.CreateModelCommand(
+        name,
+        str(model_uri),
+        schema,
+        flavor,
+        preprocessor,
```
changhiskhan, line 109: would it make sense to abstract some of the interface details behind a ModelProcessor class which could have a name and func bytes?

python/rikai/spark/model.py
```diff
@@ -0,0 +1,116 @@
...
+
+    logging.info("Register model=%s uri=%s schema=%s", name, model_uri, schema)
+
+    if options is None:
+        options = {}
+    options = {str(k): str(v) for k, v in options.items()}
+
+    preprocessor_bytes = None
+    if isinstance(preprocessor, Callable):
+        preprocessor_bytes = base64.b64encode(
```
changhiskhan, line 88: can't we just use the pickled bytes directly?

python/rikai/spark/model.py
changhiskhan: or if we can use bytes directly then the datatype would be able to tell us which one it is?

python/rikai/spark/sql/codegen/base.py
```diff
@@ -74,7 +75,9 @@ def resolve(self, spec: "ModelSpec"):
             "type": "object",
             "properties": {
                 "pre": {"type": "string"},
+                "preCode": {"type": "string"},
```
changhiskhan, line 78: if we can use bytes directly we just can just use "pre" and have a list of acceptable types?

changhiskhan (commented):
yeah, i'm a little nervous about this one. If we can some of the details before a ModelTransform class or something that probably would make it more future-proof.

python/rikai/spark/sql/codegen/base.py
eddyxu: To pass model spec from jvm to python, so we are writing the spec into a json , and let python to load it. 
That's where the `bytes` need to be encoded into string.

But it might be feasible to do base64 encode / decode right at the json serialization step. I can take a look

python/rikai/spark/model.py
eddyxu: Answered in another thread, we need to pass this from JVM to python using json. 

Probally can use other binary format, i.e., protobuf? feel a little bit overkill tho.

python/rikai/spark/model.py
eddyxu: this is calling a Scala / Java code.

The type is already explicitly in java code https://github.com/eto-ai/rikai/pull/474/files#diff-a3ca47dbaa1d8ec4cc5fcf62a11aa72ee3cc47a3d564a213fb4ea41b32ab8234R47.

I can change this field to byte[], and do the base64 encoding when passing spec from jvm into python again. 
I did not do it was because 

```
case class ModelSpec
```
currently can be easily decoded via `circe-json`. If these two fields are bytes in JVM, we need to do base64 right before `circe-json` encoding.

python/rikai/spark/model.py
changhiskhan: yeah, no need for protobuf here yet i think

PR #478, opened 2022-01-11
Support sklearn PCA models

### LINKED ISSUE #477: Support PCA model for sklearn codegen
https://github.com/eto-ai/rikai/blob/c5f32bb4580e81b75053a3be29621a7059dec9b6/python/rikai/spark/sql/codegen/sklearn.py#L41-L48

For sklearn PCA model, `model.predict` is not available.

See https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html

We need to invoke `model.transform`.

da-tubi:
closes #477 

## Why?
Because for sklearn models, there are two ways as far as I know to apply a model:
+ `model.predict(X)`
+ `model.transform(X)`

### DIFF
--- a/python/rikai/spark/sql/codegen/sklearn.py
+++ b/python/rikai/spark/sql/codegen/sklearn.py
@@ -38,13 +38,21 @@ def generate_udf(spec: "rikai.spark.sql.codegen.base.ModelSpec"):
     A Spark Pandas UDF.
     """
 
+    def predict(model, X):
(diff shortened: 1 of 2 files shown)


eddyxu:
We can probably change the title to "Support sk-learn models with either predict or transform" signature? This change is agnostic to a particular model in sklearn.

eddyxu (approved):

PR #485, opened 2022-01-12
rikai.sql.ml.registry.torchhub.enabled defaults to false

da-tubi:
Disable TorchHub Registry by default to avoid security issues.

Here is the error message with torchhub registry disabled:
```
>                   raise Py4JJavaError(
                        "An error occurred while calling {0}{1}{2}.\n".
                        format(target_id, ".", name), value)
E                   py4j.protocol.Py4JJavaError: An error occurred while calling o57.sql.
E                   : ai.eto.rikai.sql.model.ModelNotFoundException:
E                   TorchHub Registery is disabled by default for security concerns.
E                   Be cautious and set `rikai.sql.ml.registry.torchhub.enabled` to true
E                   only for personal usage or testing purpose.
E
E                   	at ai.eto.rikai.sql.model.torchhub.TorchHubRegistry.resolve(TorchHubRegistry.scala:31)
E                   	at ai.eto.rikai.sql.model.Registry$.resolve(Registry.scala:194)
```

### DIFF
--- a/src/main/scala/ai/eto/rikai/sql/model/torchhub/TorchHubRegistry.scala
+++ b/src/main/scala/ai/eto/rikai/sql/model/torchhub/TorchHubRegistry.scala
@@ -16,11 +16,25 @@
 
 package ai.eto.rikai.sql.model.torchhub
 
-import ai.eto.rikai.sql.model.PyImplRegistry
+import ai.eto.rikai.RikaiConf
+import ai.eto.rikai.sql.model.{Model, ModelNotFoundException, ModelSpec, PyImplRegistry}
+import org.apache.spark.sql.SparkSession
 
 /** TorchHub-based Model [[Registry]].
   */
 class TorchHubRegistry(val conf: Map[String, String]) extends PyImplRegistry {
+  override def resolve(session: SparkSession, spec: ModelSpec): Model = {
+    if (RikaiConf.TORCHHUB_REG_ENABLED) {
+      super.resolve(session, spec)
+    } else {
+      throw new ModelNotFoundException(message =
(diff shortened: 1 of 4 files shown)


src/main/scala/ai/eto/rikai/sql/model/torchhub/TorchHubRegistry.scala
```diff
@@ -16,11 +16,29 @@
...
+  ModelNotFoundException,
+  ModelSpec,
+  PyImplRegistry
+}
+import org.apache.spark.sql.SparkSession
 
 /** TorchHub-based Model [[Registry]].
   */
 class TorchHubRegistry(val conf: Map[String, String]) extends PyImplRegistry {
+  override def resolve(session: SparkSession, spec: ModelSpec): Model = {
```
changhiskhan, line 31: why not just remove torchhub impl from `DEFAULT_REGISTRIES` in Registry.scala? Isn't that the only thing you need to disable it by default?

darcy-shen: because here i am printing a friendly error message

darcy-shen: you suggested impl is not friendly，people will not be hinted on how to enable it

da-tubi:
As we will move TorchHub to rikai experimental, it is not necessary any more.

And RikaiConf.scala I'm using here is still valuable. I will create another pull request for collect other Rikai configs into RikaiConf.scala

PR #486, opened 2022-01-12
update mlflow due to breaking change between 1.20 -> 1.22 (read: wtf)

changhiskhan:


Renkai (approved):

darcy-shen:
which breaking changes？

changhiskhan:
i forgot exactly but it made show models fail because the underlying mlflow code changed from GET to POST (or vice versa)

PR #494, opened 2022-01-14
Fix loading models from MlflowCatalog

eddyxu:
With this PR, Rikai supports to directly run `SELECT ML_PREDICT(model, image)` over an existing model from mlflow.

### DIFF
--- a/python/rikai/spark/sql/codegen/mlflow_registry.py
+++ b/python/rikai/spark/sql/codegen/mlflow_registry.py
@@ -224,7 +224,8 @@ class MlflowRegistry(Registry):
 
     def get_options(self, spec, run):
         options = run.data.params
-        options.update(spec.get("options", {}))
+        if spec.get("options", {}):
+            options.update(spec.get("options"))
         return options
 
     def get_model_version(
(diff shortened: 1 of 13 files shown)


python/rikai/spark/sql/codegen/mlflow_registry.py
```diff
@@ -224,7 +224,8 @@ def get_model_conf(self, spec, run):
 
     def get_options(self, spec, run):
         options = run.data.params
-        options.update(spec.get("options", {}))
+        if spec.get("options", {}):
```
changhiskhan, line 227: nit: why the change here?

changhiskhan (approved):

python/rikai/spark/sql/codegen/mlflow_registry.py
eddyxu: `spec` now is passed by jvm to python for the jdbc change, which is possible to be serialized as `{"options": None}` now. 

It was originally handled by the jvm code when passing through `py4j` to change `None` to `{}`.

PR #495, opened 2022-01-15
Temporarily Disable CREATE MODEL on MlflowRegistry

eddyxu:
We've discovered that the current implement of `CREATE MODEL` using `MlflowRegistry` does not register a valid / runnable model into Mlflow. Lets disable this functionality, and find a workable design for 0.2.


eddyxu:
Filed #496 to track the improvement for 0.2

changhiskhan (approved):

PR #502, opened 2022-01-18
Fix a bug that prevents to use ML_PREDICT in multiple SparkSessions

### LINKED ISSUE #497: [BUG] Undefined function when running ML_PREDICT with a re-connected jdbc connection
### Problem:

Use `beeline` to reconnect to the ThriftServer the second time, `ML_PREDICT` can not resolve the function to render 

```
0: jdbc:hive2://localhost:10001> select ml_predict(ssd, image) as det from (select image from coco limit 1);
Error: org.apache.hive.service.cli.HiveSQLException: Error running query: org.apache.spark.sql.AnalysisException: Undefined function: 'Some(ssd)_Stream(Y, ?)_post'. This function is neither a registered temporary function nor a permanent function registered in the database 'default'.; line 1 pos 7
	at org.apache.spark.sql.hive.thriftserver.SparkExecuteStatementOperation.org$apache$spark$sql$hive$thriftserver$SparkExecuteStatementOperation$$execute(SparkExecuteStatementOperation.scala:361)
	at org.apache.spark.sql.hive.thriftserver.SparkExecuteStatementOperation$$anon$2$$anon$3.$anonfun$run$2(SparkExecuteStatementOperation.scala:263)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at org.apache.spark.sql.hive.thriftserver.SparkOperation.withLocalProperties(SparkOperation.scala:78)
	at org.apache.spark.sql.hive.thriftserver.SparkOperation.withLocalProperties$(SparkOperation.scala:62)
	at org.apache.spark.sql.hive.thriftserver.SparkExecuteStatementOperation.withLocalProperties(SparkExecuteStatementOperation.scala:43)
	at org.apache.spark.sql.hive.thriftserver.SparkExecuteStatementOperation$$anon$2$$anon$3.run(SparkExecuteStatementOperation.scala:263)
	at org.apache.spark.sql.hive.thriftserver.SparkExecuteStatementOperation$$anon$2$$anon$3.run(SparkExecuteStatementOperation.scala:258)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1730)
	at org.apache.spark.sql.hive.thriftserver.SparkExecuteStatementOperation$$anon$2.run(SparkExecuteStatementOperation.scala:272)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

The query that I run in both `beeline` sessions

```
SELECT ML_PREDICT(ssd, image) FROM (SELECT image FROM coco LIMIT 1);
```

The issue went away if i restart the spark thriftserver

### Environments

* Spark 3.1.2
* Python 3.7
* Java 1.8
* Dataset: Coco
* MlflowCatalog @ Rikai 0.0.21

eddyxu:
Closes #497 

This PR changes the model resolving from the ast parsing stage to the naming resolving stage. The bug was due to there are mulitiple spark sessions, while the JIT-ed model runner udfs were only registered to one of the SparkSession.


src/test/scala/ai/eto/rikai/sql/model/mlflow/SparkSessionWithMlflow.scala
```diff
@@ -57,7 +63,18 @@ trait SparkSessionWithMlflow extends LazyLogging with BeforeAndAfterAll {
...
 
-  private[mlflow] def clearModels(): Unit = {
+  override def beforeEach(): Unit = {
+    super.beforeEach()
+    run = mlflowClient.client.createRun()
+  }
+
+  override def afterEach(): Unit = {
+    mlflowClient.client.deleteRun(run.getRunId)
+    clearModels()
```
changhiskhan, line 73: Ok is the purpose of this test class so we can do the required setup/teardown?

src/main/scala/ai/eto/rikai/sql/spark/RikaiSparkSessionExtensions.scala
```diff
@@ -14,12 +16,90 @@
...
+}
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.rikai.expressions.{Area, IOU}
+import org.apache.spark.sql.{SparkSession, SparkSessionExtensions}
+
+private class MlPredictRule(val session: SparkSession)
+    extends Rule[LogicalPlan] {
+
+  val catalog = Catalog.getOrCreate(session)
```
changhiskhan, line 41: nit: call this modelCatalog so it's explicit (i.e., not the hive catalog) without having to look through the imports?

changhiskhan (approved):
ok so the idea is we do this during logical plan optimization phase where we can register the model udf if it's not already found?

eddyxu:
> ok so the idea is we do this during logical plan optimization phase where we can register the model udf if it's not already found?

yes, that's right. So for some reason, when starts SparkSession with the thrift server, parser / analyzer and exeuction are not necessary in the same SparkSession instance. Here, we make sure that the udfs are registered to the SparkSession which will execute the model.

A few optimizations can be done, i.e., cache the byte code of udfs and etc, to reduce the model loading time. We can put them into backlog for now.

src/test/scala/ai/eto/rikai/sql/model/mlflow/SparkSessionWithMlflow.scala
eddyxu: It clears mlflow as well as the Hive Catalog to a clean state.

src/main/scala/ai/eto/rikai/sql/spark/RikaiSparkSessionExtensions.scala
eddyxu: will do

PR #507, opened 2022-01-20
Preliminary support of tensorflow models in SQL ML

### LINKED ISSUE #145: Support tensorflow model in SQL ML
Provide a Tensorflow based batch inference runner for SQL ML.

eddyxu:
Closes #145

### DIFF
--- a/python/rikai/contrib/tensorflow/models/ssd.py
+++ b/python/rikai/contrib/tensorflow/models/ssd.py
@@ -45,7 +45,7 @@ def post_processing(options: Dict[str, Any]) -> Callable:
                     {
                         "detection_boxes": Box2d(*box),
                         "detection_classes": int(label_class),
-                        "detection_scores": score,
+                        "detection_scores": float(score),
                     }
                 )
             results.append(predict_result)
--- a/python/rikai/contrib/tensorflow/models/ssd.py
+++ b/python/rikai/contrib/tensorflow/models/ssd.py
@@ -1,0 +1,57 @@
+#  Copyright 2022 Rikai Authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""SSD: Single Shot MultiBox Detector
+
+https://arxiv.org/abs/1512.02325
+"""
+
+from typing import Any, Callable, Dict
+
+from rikai.types import Box2d
+
+HUB_URL = "https://tfhub.dev/tensorflow/ssd_mobilenet_v2/2"
+
+
+def pre_processing(options: Dict[str, Any]):
+    def pre_processor_func(batch):
+        return batch
+
+    return pre_processor_func
+
+
+def post_processing(options: Dict[str, Any]) -> Callable:
+    def post_process_func(batch):
+        results = []
+        for boxes, classes, scores in zip(
+            batch["detection_boxes"].numpy(),
+            batch["detection_classes"].numpy(),
+            batch["detection_scores"].numpy(),
+        ):
+            predict_result = []
+            for box, label_class, score in zip(boxes, classes, scores):
+                predict_result.append(
+                    {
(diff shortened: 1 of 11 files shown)


ffcai (approved):
LGTM

python/rikai/contrib/tensorflow/models/ssd.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+
+from typing import Any, Callable, Dict
+
+from rikai.types import Box2d
+
+HUB_URL = "https://tfhub.dev/tensorflow/ssd_mobilenet_v2/2"
+
+
+def pre_processing(options: Dict[str, Any]):
```
changhiskhan, line 27: should we make this a shared preprocessor so others don't need to redo it??

python/rikai/contrib/tensorflow/models/ssd.py
```diff
@@ -0,0 +1,60 @@
...
+        for boxes, classes, scores in zip(
+            batch["detection_boxes"].numpy(),
+            batch["detection_classes"].numpy(),
+            batch["detection_scores"].numpy(),
+        ):
+            predict_result = []
+            for box, label_class, score in zip(boxes, classes, scores):
+                predict_result.append(
+                    {
+                        "detection_boxes": Box2d(*box),
```
changhiskhan, line 46: would users expect the ML_PREDICT results to have a common result schema ('boxes', 'classes', 'scores') vs result schema that reflect the model/flavor being used ('detection_classes', ...)?

python/rikai/spark/sql/codegen/mlflow_logger.py
```diff
@@ -44,6 +48,37 @@ def __init__(self, flavor: str):
...
+        except ImportError as e:
+            raise ImportError(
+                "Couldn't import mlflow. Please make sure to "
+                "`pip install mlflow` explicitly or install "
+                "the correct extras like `pip install rikai[mlflow]`"
+            ) from e
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+
+            tf.saved_model.save(model, tmp_dir)
```
changhiskhan, line 71: you had to do this because the default mlflow behavior doesn't return a runnable model?

python/setup.py
```diff
@@ -25,7 +25,7 @@
 ]
 sklearn = ["scikit-learn"]
 torch = ["torch>=1.8.1", "torchvision>=0.9.1"]
-tf = ["tensorflow>=2.5.0"]
+tf = ["tensorflow>=2.5.0", "tensorflow_hub"]
```
changhiskhan, line 28: test dependency?

python/rikai/spark/sql/codegen/tensorflow.py
```diff
@@ -0,0 +1,100 @@
...
+                shape=(
+                    None,
+                    None,
+                    image_arr.shape[2],
+                ),
+                dtype=tf.uint8,
+            )
+        else:
+            raise ValueError(
+                f"Image should only have 2d or 3d shape, got {image_arr.shape}"
```
changhiskhan, line 54: do we want to support text types through this interface?

python/rikai/contrib/tensorflow/models/ssd.py
eddyxu: I would say that ML_PREDICT should returns the schema that the model defines?

```('detection_classes', ...)``` are the keys that directly returned from this particular model.

File: python/rikai/spark/sql/codegen/mlflow_logger.py (hunk 3 above)
Comment (eddyxu) [reply python/rikai/spark/sql/codegen/mlflow_logger.py:71]: Yea, the parameters of `mlflow.tensorflow.<ANYTHING>` is super weird .

For example, the first parameter of https://mlflow.org/docs/latest/python_api/mlflow.tensorflow.html#mlflow.tensorflow.log_model is a path.  I dont really know what do users expect from mlflow tbh.

File: python/rikai/spark/sql/codegen/tensorflow.py (hunk 5 above)
Comment (eddyxu) [reply python/rikai/spark/sql/codegen/tensorflow.py:54]: do you mean `text` as a new type? If so, yes. So i was thinking to put a `classmethod` to `ToNumpy` mixins, and let it return the shape of the tensor / ndarray:

```
class ToNumpy(abc.ABC):
    @abstractmethod
    def shape(self):
         pass
```

But dont have concrete idea how to do it yet. The main blocker is still the uncertainty of the shape defined in `NDArray`.

python/rikai/spark/sql/codegen/mlflow_logger.py
eddyxu: And `mlflow.tensorflow.load_model` does not actually return model, but returns model signatures 🤷

changhiskhan (approved):

PR #508, opened 2022-01-21
Ssd

changhiskhan:
create separate transforms for ssd because it needs resize


changhiskhan:
oh that reminds me @eddyxu does the TF SSD model also need the resize?

eddyxu (approved):

eddyxu:
For this particular one, https://tfhub.dev/tensorflow/ssd_mobilenet_v2/2 it does not, because it does not support batch. This model do resize within the model.

PR #512, opened 2022-01-24
Improve default GPU behavior in Torch runner

eddyxu:


python/rikai/spark/sql/codegen/pytorch.py
```diff
@@ -84,6 +85,13 @@ def torch_inference_udf(
                     results.extend(bin_predictions)
                 yield pd.Series(results)
 
+        # Release GPU memory
+        # https://blog.paperspace.com/pytorch-memory-multi-gpu-debugging/?ref=tfrecipes
+        if use_gpu:
```
changhiskhan, line 90: does this need to go in a finally block?

changhiskhan (approved):

python/rikai/spark/sql/codegen/pytorch.py
eddyxu: Good point. it is always better to do so.

PR #514, opened 2022-01-25
Fix rtd build with missing dependency

eddyxu:


changhiskhan (approved):
thanks!

eddyxu:
Tested with:

```
conda create -n rikai_docs python=3.9
conda activate rikai_docs
pip install -r python/deps/requirements_rtd.txt
cd docs
make html
```

Issue #421, opened 2021-12-28
Support Apache Spark 3.2.0


darcy-shen:
For the Scala side, #337 had made it work for Spark 3.2.0.

For the Python side, we still need to make minor changes.

## Watch
+ [x] Databricks Runtime 10.x: https://docs.databricks.com/release-notes/runtime/10.0.html
+ [ ] AWS EMR: https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-release-components.html
+ [ ] DataProc: https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-versions

Renkai:
I will work on the Python compatibility of spark 3.2.0

PR #531, opened 2022-01-30
Python Support for Spark 3.2.0

Renkai:
This patch closes https://github.com/eto-ai/rikai/issues/421

Pyspark only supports Scala 2.12, so this patch checks the Spark version to define macros, not assume Scala 2.12 will use Spark 3.1.x and Scala 2.13 will use Spark 3.2.x

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -37,6 +37,7 @@ jobs:
     - name: sbt install
       run: |
         echo "SPARK_VERSION=${{matrix.spark-version}}" >> $GITHUB_ENV
+        export SPARK_VERSION=${{matrix.spark-version}}
         sbt publishLocal
     - name: Start docker-compose
       run: |
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -46,7 +46,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: /opt/hostedtoolcache/rikai/${{ matrix.python-version }}
-        key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ hashFiles('**/setup.py') }}
+        key: ${{ runner.os }}-pip-${{ matrix.spark-version }}-${{ hashFiles('**/setup.py') }}
         restore-keys: |
           ${{ runner.os }}-pip-
     - name: apt update and install
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -12,6 +12,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7, 3.8, 3.9]
+        spark-version: [3.1.2, 3.2.0]
     timeout-minutes: 30
     steps:
     - uses: actions/checkout@v2
@@ -35,6 +36,7 @@ jobs:
         key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
     - name: sbt install
       run: |
+        echo "SPARK_VERSION=${{matrix.spark-version}}" >> $GITHUB_ENV
         sbt publishLocal
     - name: Start docker-compose
       run: |
--- a/Makefile
+++ b/Makefile
@@ -36,15 +36,14 @@ python/rikai/spark/sql/generated/RikaiModelSchemaParser.py: src/main/antlr4/org/
 
 lint:
 	sbt scalafmtCheckAll
-	black -l 79 --check python contrib/
+	black --check python contrib experimental
 	pycodestyle python contrib/
 .PHONY: lint
 
 # Fix code style
 fix:
 	sbt scalafmtAll
-	isort python contrib experimental
-	black -l 79 python contrib experimental
+	black python contrib experimental
 .PHONY: fix
 
 # increment to the next released version and add a release tag
--- a/build.sbt
+++ b/build.sbt
@@ -52,8 +52,15 @@ scmInfo := Some(
 )
 
 libraryDependencies ++= {
-  val sparkVersion =
-    if (scalaVersion.value.compareTo("2.12.15") >= 0) "3.2.0" else "3.1.2"
+  val sparkVersion = {
+    sys.env.get("SPARK_VERSION") match {
+      case Some(sparkVersion) => sparkVersion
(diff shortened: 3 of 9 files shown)


build.sbt
```diff
@@ -52,8 +52,15 @@ scmInfo := Some(
 )
 
 libraryDependencies ++= {
-  val sparkVersion =
-    if (scalaVersion.value.compareTo("2.12.15") >= 0) "3.2.0" else "3.1.2"
+  val sparkVersion = {
+    sys.env.get("SPARK_VERSION") match {
```
darcy-shen, line 56: nit: use getOrElse

darcy-shen (approved):
LGTM

Makefile
```diff
@@ -36,15 +36,14 @@ python/rikai/spark/sql/generated/RikaiModelSchemaParser.py: src/main/antlr4/org/
...
 	sbt scalafmtCheckAll
-	black -l 79 --check python contrib/
+	black --check python contrib experimental
 	pycodestyle python contrib/
 .PHONY: lint
 
 # Fix code style
 fix:
 	sbt scalafmtAll
-	isort python contrib experimental
```
eddyxu, line 46: Why remove `isort` from `make fix`?

python/setup.py
```diff
@@ -13,7 +14,7 @@
 
 # extras
 dev = [
-    "black",
+    "black==21.11b1",
```
eddyxu, line 17: the last stable one is 22.10?

https://pypi.org/project/black/

python/setup.py
```diff
@@ -68,7 +73,7 @@
         "pandas",
         "Pillow",
         "pyarrow>=6.0",
-        "pyspark==3.1.2",
+        f"pyspark=={spark_version}",
```
eddyxu, line 76: When we do a rikai release, what version should we package? Do we need to release two versions of rikai , each for one spark version?

if the python code base can work with both spark 3.1 and 3.2, should we just make this `pyspark>=3.1,<3.3`?

src/main/scala/ai/eto/rikai/sql/spark/execution/ModelCommand.scala
```diff
@@ -20,17 +20,26 @@ import ai.eto.rikai.sql.model.Catalog
...
+  )
   override final def mapChildren(f: LogicalPlan => LogicalPlan): LogicalPlan =
     this.asInstanceOf[LogicalPlan]
-  @enableIf(scala.util.Properties.versionNumberString.compareTo("2.12.15") >= 0)
+
+  @enableIf(c =>
+    c.classPath.exists(
+      _.getPath.matches(".*spark-catalyst_2\\.\\d+-3\\.2\\..*")
+    )
+  )
```
eddyxu, line 42: nice,
what will we do for Spark 3.3 then ;)

src/main/scala/org/apache/spark/sql/rikai/expressions/Box2d.scala
```diff
@@ -45,7 +45,11 @@ case class Area(child: Expression)
 
   override def prettyName: String = "area"
 
-  @enableIf(scala.util.Properties.versionNumberString.compareTo("2.12.15") >= 0)
+  @enableIf(c =>
+    c.classPath.exists(
+      _.getPath.matches(".*spark-catalyst_2\\.\\d+-3\\.2\\..*")
```
eddyxu, line 50: These rules seems to be the same? shall we just extract them into one place, like `object SparkVersion` or something?

python/setup.py
eddyxu: One more questions, is the scala code `enableIf` built by compile time or runtime?
So if user install rikai, do they need to specify different jar version ?

Makefile
Renkai: I found out `profile = black` for isort seems only works well in simple conditions, but we [customized black](https://github.com/eto-ai/rikai/blob/main/pyproject.toml#L1) a lot, so they just can't be compatible with each other and I can't find out what leads to it. May be use `black` as a forced standard in CI and make `isort` optional is better.

python/setup.py
Renkai: It makes sense, I will change to `22.1.0`

python/setup.py
Renkai: `enableIf` works in compile-time. I'm afraid we have to release `rikai_spark31_2.12`, `rikai_spark32_2.12`, `rikai_spark32_2.13` separately, that's the pain of building a system on top of another big system.

src/main/scala/ai/eto/rikai/sql/spark/execution/ModelCommand.scala
Renkai: We can use `_.getPath.matches(".*spark-catalyst_2\\.\\d+-3\\.(2|3)\\..*")` or similar things, `regex` is very powerful, just will become less readable as it become more powerful.

src/main/scala/org/apache/spark/sql/rikai/expressions/Box2d.scala
Renkai: It's a compile-time plugin, I'm not sure if it can access the source code when it's compiling. A practical way is to put this code into a submodule and to make the submodule a compiler plugin of other modules, which will increase the complexity of build-time work, and may also let IDEs in trouble.

Renkai:
This patch would add Spark 3.2 to CI, but would not release a public version that supports Spark 3.2.
We can change the release script later when people really need rikai for Spark 3.2

.github/workflows/python.yml
```diff
@@ -43,7 +46,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: /opt/hostedtoolcache/rikai/${{ matrix.python-version }}
-        key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ hashFiles('**/setup.py') }}
+        key: ${{ runner.os }}-pip-${{ matrix.spark-version }}-${{ hashFiles('**/setup.py') }}
```
darcy-shen, line 49: why the key is determined only by `spark-version`?

PR #542, opened 2022-02-16
use Model Type for TorchHub Registry

da-tubi:
+ Deprecate pre/post/OUTPUT_SCHEMA support for TorchHub Registry
+ add Model Type support for TorchHub Registry

Issue #533, opened 2022-01-30
Provide utilities to define model types


eddyxu:
Improve api / developer experience to build new model types.

darcy-shen:
The Model Type approach is much better. For me, I need utilities to test model types without Spark.

darcy-shen:
Test MODEL_TYPE without Spark Session:
https://github.com/eto-ai/rikai/pull/546

PR #544, opened 2022-02-16
Unit Test Draft on Model Type

da-tubi:
This is a draft PR waiting for @eddyxu 's feedback.

This first three commits are from #542 , the last two commits belong to this PR.

How to run the unit test on Model Type:
``` bash
cd experimental/torchhub/
pytest -s -o log_cli=true tests/test_resnet.py -k test_resnet_model_type
```

### DIFF
--- a/experimental/torchhub/tests/test_resnet.py
+++ b/experimental/torchhub/tests/test_resnet.py
@@ -13,18 +13,43 @@
 #  limitations under the License.
 
 from pathlib import Path
+from rikai.types.vision import Image
 
 import torchvision
+import pandas as pd
 from pyspark.sql import SparkSession
 
 from rikai.spark.functions import to_image
+from rikai.experimental.torchhub.torchhub_registry import (
+    TorchHubRegistry,
+)
+from rikai.spark.sql.codegen.base import func_from_spec
+
+
+version = f"v{torchvision.__version__.split('+', maxsplit=1)[0]}"
+work_dir = Path().absolute().parent.parent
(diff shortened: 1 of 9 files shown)


python/rikai/spark/sql/codegen/pytorch.py
```diff
@@ -41,7 +41,7 @@ def move_tensor_to_device(data, device):
     return data
 
 
-def generate_udf(payload: ModelSpec):
+def generate(payload: ModelSpec, is_udf: bool = True):
```
darcy-shen, line 44: Use is_udf = False to generate the function which can be used without PySpark

only for pytorch now, need feedbacks from reviewers and then apply to other codegen

experimental/torchhub/tests/test_resnet.py
```diff
@@ -13,18 +13,43 @@
...
+)
+from rikai.spark.sql.codegen.base import func_from_spec
+
+
+version = f"v{torchvision.__version__.split('+', maxsplit=1)[0]}"
+work_dir = Path().absolute().parent.parent
+image_path = f"{work_dir}/python/tests/assets/test_image.jpg"
+
+
+def test_resnet_model_type():
```
darcy-shen, line 34: This is an attempt to provide mode type utilities. It aims to provide a pyspark-free way to test the model type.

See https://github.com/eto-ai/rikai/issues/533

darcy-shen:
will create another pr rebased on main

Issue #539, opened 2022-02-15
Torch is required even when we are using Tensorflow


darcy-shen:
See https://github.com/eto-ai/rikai/runs/5199485143?check_suite_focus=true
from #538 

```
[...]

Issue #40, opened 2021-01-23
Support tensorflow Dataset 


eddyxu:
Support reading rikai dataset in Tensorflow

Renkai:
Solved by
https://github.com/eto-ai/rikai/pull/574

PR #553, opened 2022-02-18
Tensor Dataset support

Renkai:
This patch solves
https://github.com/eto-ai/rikai/issues/539 
and 
https://github.com/eto-ai/rikai/issues/40

### DIFF
--- a/python/rikai/spark/sql/codegen/tensorflow.py
+++ b/python/rikai/spark/sql/codegen/tensorflow.py
@@ -78,7 +78,7 @@ def _generate(payload: ModelSpec, is_udf: bool = True):
 
             data = PandasDataset(
                 df, model.transform(), unpickle=is_udf, use_pil=True
-            ).data()
+            ).data(batch_size)
 
             results = []
             for batch in data:
--- a/python/rikai/spark/sql/codegen/tensorflow.py
+++ b/python/rikai/spark/sql/codegen/tensorflow.py
@@ -20,8 +20,8 @@ from pyspark.serializers import CloudPickleSerializer
 from pyspark.sql.functions import pandas_udf
 from pyspark.sql.types import BinaryType
 
-from rikai.pytorch.pandas import PandasDataset
 from rikai.spark.sql.model import ModelSpec
+from rikai.tensorflow.pandas import PandasDataset
 from rikai.types import Image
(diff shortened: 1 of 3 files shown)


python/rikai/tensorflow/pandas.py
```diff
@@ -0,0 +1,67 @@
...
+    ) -> None:
+        assert isinstance(df, (pd.DataFrame, pd.Series))
+        self.df = df
+        self.transform = transform
+        self.unpickle = unpickle
+        self.use_pil = use_pil
+
+    def data(self, batch_size):
+
+        arr = self.df[0]
```
Renkai, line 57: From current tests, there is always only one item in `self.df`, not sure if it's the designed condition.

darcy-shen: we could double and double the dataset, and then, log the size of the pandas udf input

da-tubi: you can use https://github.com/eto-ai/rikai/blob/v0.1.4/experimental/tfhub/tests/test_tfhub.py to run the test

darcy-shen: you can use https://github.com/eto-ai/rikai/blob/v0.1.4/experimental/tfhub/tests/test_tfhub.py to run the test

python/rikai/spark/sql/codegen/tensorflow.py
```diff
@@ -76,17 +76,9 @@ def tf_inference_udf(
...
-                    shape=(None, None, 3), dtype=tf.uint8, name="input_tensor"
-                ),
-            )
-
-            if model.transform():
-                data = data.map(model.transform())
-            data = data.batch(batch_size)
+            data = PandasDataset(
+                df, model.transform(), unpickle=is_udf, use_pil=True
+            ).data(batch_size)
```
eddyxu, line 81: `.data => .batch`?

python/rikai/tensorflow/pandas.py
eddyxu: this does not sound right. can we check batch size and etc. ?

Renkai:
More complex than I thought, will create another branch to avoid noise during debugging.

Issue #332, opened 2021-10-13
UDT: Can't extract value from image#98: need struct type but got struct<data:binary,uri:string>


darcy-shen:
``` python
df.selectExpr("image.uri")
```

```
---------------------------------------------------------------------------
AnalysisException                         Traceback (most recent call last)
/var/folders/6d/q8_1m19n1jzg6_3lghwylhpc0000gp/T/ipykernel_82167/3230124512.py in <module>
----> 1 df.selectExpr("image.uri")

~/.pyenv/versions/3.8.10/envs/rikai-example/lib/python3.8/site-packages/pyspark/sql/dataframe.py in selectExpr(self, *expr)
   1684         if len(expr) == 1 and isinstance(expr[0], list):
   1685             expr = expr[0]
-> 1686         jdf = self._jdf.selectExpr(self._jseq(expr))
   1687         return DataFrame(jdf, self.sql_ctx)
   1688 

~/.pyenv/versions/3.8.10/envs/rikai-example/lib/python3.8/site-packages/py4j/java_gateway.py in __call__(self, *args)
   1302 
   1303         answer = self.gateway_client.send_command(command)
-> 1304         return_value = get_return_value(
   1305             answer, self.gateway_client, self.target_id, self.name)
   1306 

~/.pyenv/versions/3.8.10/envs/rikai-example/lib/python3.8/site-packages/pyspark/sql/utils.py in deco(*a, **kw)
    115                 # Hide where the exception came from that shows a non-Pythonic
    116                 # JVM exception message.
--> 117                 raise converted from None
    118             else:
    119                 raise

AnalysisException: Can't extract value from image#98: need struct type but got struct<data:binary,uri:string>; line 1 pos 0
```

eddyxu:
TODO: explore https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/types/SQLUserDefinedType.html

PR #565, opened 2022-03-08
Implement to_struct function to convert UDT to Struct

eddyxu:
```
SELECT to_struct(image).uri FROM images

SELECT to_struct(box) as bigbox FROM boxes WHERE to_struct(box).xmin > 3
```

Partially mitigate #332.


eddyxu:
It is difficult to make a `SparkSessionExtensions` to ingest a rule to handle this at the moment.
`SparkSessionExtensions.ingestResolutionRule` happens AFTER spark sql parser raises the error of #332 .

To completely fix this, we should do it in upstream:

https://github.com/apache/spark/blob/71991f75ff441e80a52cb71f66f46bfebdb05671/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/complexTypeExtractors.scala#L52-L71

darcy-shen (approved):
Nice workaround!

Issue #560, opened 2022-03-02
Model Type Design: optional uri


darcy-shen:
``` sql
CREATE (OR REPLACE)? MODEL (IF NOT EXISTS)? model=qualifiedName
      (FLAVOR flavor=identifier)?
      (MODEL_TYPE modeltype=qualifiedName)?
      (OPTIONS optionList)?
      (RETURNS datatype=dataType)?
      (USING uri=STRING)
```

URI specifies how to load a model. Sometimes, we load a model in that way:

``` python
import ml_pkg

model = ml_pkg.ocr_model()
preds = model(inputs)
```
For the above use case, we do not need a model registry (or we need an anonymous model registry).

BootstrapModelType is also needed to verify if a ModelType can be bootstrapped.

PR #569, opened 2022-03-09
Support Bootstrappable ModelType

da-tubi:
## Main Feature
No need to depend on extra model registry
``` sql
-- before
CREATE MODEL ssd_model
MODEL_TYPE ssd
USING 'mlflow://ssd'

-- now
CREATE MODEL ssd_model FLAVOR pytorch MODEL_TYPE ssd
```

Just make your ModelType bootstrappable, and provide the bootstrap method to load the model.
``` python
+    def bootstrappable(self) -> bool:
+       return True
+
+    def bootstrap(self):
+        return torchvision.models.detection.fasterrcnn_resnet50_fpn()
```

## Extra Bennefits
For model service (eg. [AWS rekognition](https://aws.amazon.com/rekognition/)), we do not load a model, we just call the API and get the result.

### DIFF
--- a/python/rikai/pytorch/models/__init__.py
+++ b/python/rikai/pytorch/models/__init__.py
@@ -19,6 +19,7 @@ from typing import Optional
 import torch
 
 from rikai.spark.sql.model import ModelSpec, ModelType
+from rikai.spark.sql.codegen.bootstrap import BootstrapModelSpec
 
 __all__ = ["TorchModelType"]
 
@@ -31,7 +32,13 @@ class TorchModelType(ModelType, ABC):
         self.spec: Optional[ModelSpec] = None
 
     def load_model(self, spec: ModelSpec, **kwargs):
-        self.model = spec.load_model()
+        if isinstance(spec, BootstrapModelSpec):
+            if self.bootstrappable():
+                self.model = self.bootstrap()
+            else:
+                raise RuntimeError("ModelType is not bootstrappable")
+        else:
+            self.model = spec.load_model()
         self.model.eval()
         if "device" in kwargs:
             self.model.to(kwargs.get("device"))
--- a/python/rikai/pytorch/models/fasterrcnn_resnet50_fpn.py
+++ b/python/rikai/pytorch/models/fasterrcnn_resnet50_fpn.py
@@ -13,11 +13,18 @@
 #  limitations under the License.
 
 from .torchvision import ObjectDetectionModelType
+import torchvision
 
 __all__ = ["MODEL_TYPE"]
 
 
 class FasterRCNNModelType(ObjectDetectionModelType):
+    def bootstrappable(self) -> bool:
+        return True
+
+    def bootstrap(self):
+        return torchvision.models.detection.fasterrcnn_resnet50_fpn()
+
     def __init__(self):
         super().__init__("fasterrcnn_resnet50_fpn")
 
--- a/python/rikai/pytorch/models/ssd.py
+++ b/python/rikai/pytorch/models/ssd.py
@@ -13,11 +13,18 @@
 #  limitations under the License.
 
 from .torchvision import ObjectDetectionModelType
+import torchvision
 
 __all__ = ["MODEL_TYPE"]
 
 
 class SSDModelType(ObjectDetectionModelType):
+    def bootstrappable(self) -> bool:
+        return True
+
+    def bootstrap(self):
+        return torchvision.models.detection.ssd.ssd300_vgg16()
+
     def __init__(self):
         super().__init__("SSD")
 
--- a/python/rikai/spark/sql/codegen/bootstrap.py
+++ b/python/rikai/spark/sql/codegen/bootstrap.py
@@ -21,45 +21,22 @@ __all__ = ["BootstrapRegistry"]
 
 
 class BootstrapModelSpec(ModelSpec):
-    """Bootstraped Model Spec.
-
-    Parameters
-    ----------
-    options : Dict[str, Any], optional
-        Additionally options. If the same option exists in spec already,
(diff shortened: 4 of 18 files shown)


eddyxu:
We already have `tfhub` and `tensorhub` supports, this PR seems to be an even smaller corner case while adds extra complexity for the adoption, isn't it?

Loading pre-trained models does not benefit the majority of the end-users. It only benefits to the people who is developing the model types, IIUC?

Let's spend sometime to make users happy first?

da-tubi:
> We already have tfhub and tensorhub supports, this PR seems to be an even smaller corner case while adds extra complexity for the adoption, isn't it?

No. 

All other registries could be implemented in BootstrapRegistry with `bootstrappable=True` and `bootstrap` provided. So it is not smaller corner case.

> Loading pre-trained models does not benefit the majority of the end-users. It only benefits to the people who is developing the model types, IIUC?

It is for:
1. Developing the model type
2. Wrap model service in ModelType (for this service, there is no way to load the model)
     + AWS rekognition
     + [easyocr](https://github.com/JaidedAI/EasyOCR)  https://github.com/darcy-shen/rikai-easyocr/wiki

ffcai:
> 

I talked with some engineers in perception team yesterday, they care more about the flexibility of developing the models than running the models and the deployment. I even spent some time to persuade them to not write UDF by their own.  I think developing the model type is not a corner use case. e.g. we are going to customize the mmdetection model type to enable batch inference, meanwhile we don't want to bind with tfhub or mlflow too early.

eddyxu:
For the second case 

```Wrap model service in ModelType (for this service, there is no way to load the model) AWS rekognition```

You dont really need `bootstrap` a model in this case, isnt it?  You just need to initialize a `boto` client in the constructor, and skip implementing `load_model`.  `ModelType.predict()` can call the `boto` Client directly.

And for other models of which a concrete model is needed (i.e., `pytorch`,  `tensorflow`, `sklearn`). You can just offer a default model in the constructor or lazily initialize a model if model URL is not presented, without introducing new interface. 

I'd lean to keep the interface as small as possible. The more interface we add, the less likely this project will be adopted easily by users later.

@ffcai 

> we are going to customize the mmdetection model type to enable batch inference, meanwhile we don't want to bind with tfhub or mlflow too early.

Curious, does your team expect each model always return a pre-trained / fixed model when developing a model type? Or do they need to run different trained models from the same architecture? if so, an extra identity other than `ModelType` is needed.

**Comment by da-tubi** [comment]
> And for other models of which a concrete model is needed (i.e., pytorch, tensorflow, sklearn). You can just offer a default model in the constructor or lazily initialize a model if model URL is not presented, without introducing new interface.

ModelType which is bootstrappable can be used by all registries.

> You can just offer a default model in the constructor or lazily initialize a model if model URL is not presented, without introducing new interface.

This new interface is really simple. Just add two methods.

> I'd lean to keep the interface as small as possible. The more interface we add, the less likely this project will be adopted easily by users later.

The two extra interface is optional. For normal use case, we do not have to use it.

**Comment by da-tubi** [comment]
> You dont really need bootstrap a model in this case, isnt it? You just need to initialize a boto client in the constructor, and skip implementing load_model. ModelType.predict() can call the boto Client directly.

How about just adding one extra interface?

``` python
def bootstrappable()
```

For BootstrapRegistry, we need at least one extra method to indicate whether a model is bootstrappable.

Because bootstrappable ModelType can be used by all Rikai registries. It is not a good idea to add a python class like BootstrapModelType and let all bootstrappable ModelType extend BootstrapModelType.

da-tubi:
@eddyxu  What's your idea on the BootstrapRegistry?

If you do not accept adding new interface to ModelType, can we still implement BootstrapRegistry?

For example, we can put BootstrapRegistry in `rikai.experimental`. And if a ModelType is not bootstrappable, just let it crash.

eddyxu:
I think i am still not convinced that why do we need `bootstrappable` interface as well as `BootstrapRegistry`? 

I am not in favor of adding these concepts to address small use cases, while it is achievable by other means. I'd rather to have 1 simple way to do multiple cases instead each case has one collection of concepts.  Especially these interfaces are public facing (not just for Rikai developers).

Is this design for the case in #560 (Optional URI)?

If so, we can just handle this problem in [load_model](https://github.com/eto-ai/rikai/blob/main/python/rikai/spark/sql/model.py#L207-L209)?
And with that, we dont need a model registry right? because `URI` does not exist at all?

Also is `bootstrap` a well-known term to ML engineers in the sense of `get pre-trained model by default`?

da-tubi:
> Is this design for the case in https://github.com/eto-ai/rikai/issues/560 (Optional URI)?

Yes.

> If so, we can just handle this problem in [load_model](https://github.com/eto-ai/rikai/blob/main/python/rikai/spark/sql/model.py#L207-L209)?

I think No. Could you describe in detail on how to handle this problem in load_model.

For example, with this PR, the only thing I need to change for fasterrcnn_resnet50_fpn is:

```
+    def bootstrappable(self) -> bool:
+       return True
+
+    def bootstrap(self):
+        return torchvision.models.detection.fasterrcnn_resnet50_fpn()
```

How to implement without applying this PR (using Rikai 0.1.4 or applying another better PR)?


> And with that, we dont need a model registry right? because URI does not exist at all?

When URI does not exist at all, I think a dummy registry (BootstrapRegistry in this case) is needed.
With a dummy registry, we can make minimum changes to the current impl.

> Also is bootstrap a well-known term to ML engineers in the sense of get pre-trained model by default?

Welcome to suggest a better name.

da-tubi:
plan to use only one extra method `pretrained` in ModelType, closing this PR

PR #574, opened 2022-03-11
Tensor dataset support2

Renkai:
@darcy-shen advised adding some tests to verify the TensorFlow dataset works fine both in a single input picture and muti-pictures.

I added the tests and finally made them pass, but I still have something not intelligible for me.
One thing is 
we can't use `tensors = self.df.map(upickle_convent_transform).to_numpy()` but we can use `tensors = np.array([upickle_convent_transform(x) for x in self.df])`, I believe the former one is more efficient, but tensorflow refused to use it to construct a dataset.

Another thing is, the SSD model only accept data with shape `(1,x,x,3)`, which means we can only pass in one picture into that model, but the picture has to be wrapped in an array with length 1, I can't understand why it's signature should be like that.

I would remove the `print`s after other parts look good to you.

### DIFF
--- a/experimental/tfhub/tests/test_tfhub.py
+++ b/experimental/tfhub/tests/test_tfhub.py
@@ -69,7 +69,6 @@ def test_ssd_model_type2():
 
 
 def test_ssd(spark: SparkSession):
-    spark.udf.register("to_image", to_image)
     spark.sql(
         f"""
         CREATE MODEL tfssd
@@ -87,7 +86,6 @@ def test_ssd(spark: SparkSession):
 
 
 def test_multi_pics_ssd(spark: SparkSession):
-    spark.udf.register("to_image2", to_image)
     spark.sql(
         f"""
         CREATE MODEL tfssd2
@@ -98,7 +96,7 @@ def test_multi_pics_ssd(spark: SparkSession):
     )
 
     spark.range(10).selectExpr(
-        "id as id", f"to_image2('{image_path}') as image"
+        "id as id", f"to_image('{image_path}') as image"
     ).createOrReplaceTempView("test_view")
 
     result = spark.sql(
--- a/experimental/tfhub/tests/test_tfhub.py
+++ b/experimental/tfhub/tests/test_tfhub.py
@@ -53,7 +53,7 @@ def test_ssd_model_type2():
             ]
         )
     ]
-    results_list = apply_model_spec(
+    results_iter = apply_model_spec(
         {
             "name": "tfssd",
             "uri": f"tfhub:///tensorflow/ssd_mobilenet_v2/2",
@@ -61,8 +61,10 @@ def test_ssd_model_type2():
         },
         inputs_list,
     )
-    assert len(results_list) == 1
-    series = results_list[0]
+
+    assert len(results_iter) == 1
+    series = results_iter[0]
+    assert series.shape[0] == 6
     assert len(series[0]) == 100
 
 
--- a/experimental/tfhub/tests/test_tfhub.py
+++ b/experimental/tfhub/tests/test_tfhub.py
@@ -73,10 +73,10 @@ def test_ssd(spark: SparkSession):
 
 
 def test_multi_pics_ssd(spark: SparkSession):
-    spark.udf.register("to_image", to_image)
+    spark.udf.register("to_image2", to_image)
     spark.sql(
         f"""
-        CREATE MODEL tfssd
+        CREATE MODEL tfssd2
         MODEL_TYPE ssd
         OPTIONS (device="cpu", batch_size=32)
         USING "tfhub:///tensorflow/ssd_mobilenet_v2/2";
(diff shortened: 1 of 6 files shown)


experimental/tfhub/tests/test_tfhub.py
```diff
@@ -40,6 +40,32 @@ def test_ssd_model_type():
...
+        {
+            "name": "tfssd",
+            "uri": f"tfhub:///tensorflow/ssd_mobilenet_v2/2",
+            "modelType": "ssd",
+        },
+        inputs_list,
+    )
+    assert len(results_list) == 1
+    series = results_list[0]
+    assert len(series[0]) == 100
```
darcy-shen, line 66: we have 6 images, adding assertion for the number 6?

experimental/tfhub/tests/test_tfhub.py
```diff
@@ -56,3 +82,27 @@ def test_ssd(spark: SparkSession):
     """
     )
     assert result.count() == 1
+
+
+def test_multi_pics_ssd(spark: SparkSession):
+    spark.udf.register("to_image2", to_image)
```
darcy-shen, line 88: why we need to_image2?

darcy-shen: nit: how about register all udfs or the used udfs here

https://github.com/eto-ai/rikai/blob/fceff116560dc16c1f588cf25ea208fe0b1220e6/experimental/tfhub/tests/conftest.py#L22

python/rikai/contrib/tfhub/tensorflow/ssd.py
```diff
@@ -37,27 +36,33 @@ def transform(self) -> Callable:
         return None
 
     def predict(self, images, *args, **kwargs) -> Any:
+        print("images shape", images.shape)
+        # one image should have shape (height,width,3),
```
darcy-shen, line 40: This model does not support batching as stated here:

https://tfhub.dev/tensorflow/ssd_mobilenet_v2/2

> A three-channel image of variable size - the model does NOT support batching. The input tensor is a tf.uint8 tensor with shape [1, height, width, 3] with values in [0, 255].

I suggest that you should add some comments here.

experimental/tfhub/tests/test_tfhub.py
```diff
@@ -56,3 +82,27 @@ def test_ssd(spark: SparkSession):
...
+    spark.sql(
+        f"""
+        CREATE MODEL tfssd2
+        MODEL_TYPE ssd
+        OPTIONS (device="cpu", batch_size=32)
+        USING "tfhub:///tensorflow/ssd_mobilenet_v2/2";
```
darcy-shen, line 94: Here is the corresponding TFHUB uri: https://tfhub.dev/tensorflow/ssd_mobilenet_v2/2

experimental/tfhub/tests/test_tfhub.py
Renkai: I found it's actually an array of one `Series`, and the `Series` has a shape `(6,...)`.
I updated the assertion, notify me if it's not expected.

experimental/tfhub/tests/test_tfhub.py
Renkai: updated

python/rikai/contrib/tfhub/tensorflow/ssd.py
Renkai: updated, do you have any new ideas here? https://github.com/eto-ai/rikai/pull/574/files#diff-0ee92fc67483ddd537672653ff62e0da925334688d1a92cc5c1166d5dec840f7R76

python/rikai/tensorflow/pandas.py
```diff
@@ -0,0 +1,76 @@
...
+        use_pil: bool = False,
+    ) -> None:
+        assert isinstance(df, pd.Series)
+        self.df = df
+        self.transform = transform
+        self.unpickle = unpickle
+        self.use_pil = use_pil
+
+    def batch(self, batch_size):
+        def upickle_convent_transform(entity):
```
darcy-shen, line 55: typo: convent -> convert

python/rikai/tensorflow/pandas.py
```diff
@@ -0,0 +1,76 @@
...
+            if self.unpickle:
+                entity = unpickle_transform(entity)
+            entity = convert_tensor(entity, use_pil=self.use_pil)
+
+            from rikai.types.vision import Image
+
+            ret = Image.from_pil(entity).to_numpy()
+            if self.transform:
+                ret = self.transform(ret)
```
darcy-shen, line 64: See https://github.com/eto-ai/rikai/blob/7051e5c7a0951780481cdee657444d98337fcf2a/python/rikai/pytorch/pandas.py#L60-L62

The PyTorch Pandas Dataset is using PIL as the default input type for transform. But I think it is ok to default to ndarray here. I think @eddyxu will improve the interface here later as we discussed last time.

darcy-shen (approved):
LGTM, I think it is good enough as the second attempt.

Issue #578, opened 2022-03-18
Support AWS Rekognition


darcy-shen:
AWS Rekognition: https://aws.amazon.com/rekognition/

Issue #564, opened 2022-03-07
New Model Flavor: PaddlePaddle


darcy-shen:
+ MLflow support: https://www.mlflow.org/docs/latest/python_api/mlflow.paddle.html
+ https://github.com/PaddlePaddle/PaddleOCR
+ https://github.com/PaddlePaddle/PaddleDetection
+ https://github.com/PaddlePaddle/PaddleVideo

PR #579, opened 2022-03-21
Model Type with Pretrained Mixin

### LINKED ISSUE #560: Model Type Design: optional uri

da-tubi:
Second attempt to close #560 

Pretrained Model Type is designed for the development of ModelType. Just like TFHub and TorchHub support, it is to make Rikai much more user-friendly. The difference is:

+ TFHub/TorchHub is designed for first-time Rikai SQL users
+ Pretrained Model Type is for Rikai users who cares about both SQL and Python

There are several different ways to load a pretrained model:
1. The cannonical way: eg. TFHub/TorchHub `hub.load('path/to/model')`
2. The python way: eg. `torchvision.models.detection.fasterrcnn_resnet50_fpn()`
3. Others: eg. `torch.load("https://path/to/model.pt")`
4. onnx: eg. `onnx.load("xxx.onnx")`
5. More to come

There is already a nice abstraction for the first case. For the other cases, pretrained ModelType could and should be used.

Well, with pretrained ModelType, we might need a simplest baseline/dummy flavor for #578 .


darcy-shen:
With this PR merged, we do not need to add MLflow support when working on #580 and #564 .

python/rikai/pytorch/models/fasterrcnn_resnet50_fpn.py
```diff
@@ -12,12 +12,28 @@
...
 __all__ = ["MODEL_TYPE"]
 
 
-class FasterRCNNModelType(ObjectDetectionModelType):
+class FasterRCNNModelType(ObjectDetectionModelType, Pretrained):
+    def load_pretrained_model(self):
+        return torchvision.models.detection.fasterrcnn_resnet50_fpn()
+
+    def load_model(self, spec: ModelSpec, **kwargs):
+        if isinstance(spec, DummyModelSpec):
```
eddyxu, line 30: Can we do this dispatch in Rikai's model loader code instead of user-facing ModelType?

src/main/scala/ai/eto/rikai/sql/model/dummy/DummyRegistry.scala
```diff
@@ -0,0 +1,12 @@
+package ai.eto.rikai.sql.model.dummy
+
+import ai.eto.rikai.sql.model.PyImplRegistry
+import com.typesafe.scalalogging.LazyLogging
+
+/** NoURIRegistry is used when no registry URI is specified
```
eddyxu, line 6: NoURIRegistry => DummyRegistry.

Please add license as well

python/rikai/mixin.py
```diff
@@ -126,3 +133,11 @@ def open(self, mode="rb") -> BinaryIO:
         if self.is_embedded:
             return BytesIO(self.data)
         return open_uri(self.uri, mode=mode)
+
+
+class Pretrained(ABC):
+    """Mixin for pretrained model"""
+
+    @abstractmethod
+    def load_pretrained_model(self) -> Any:
```
eddyxu, line 142: Given that this is the only method of pretrained Mixin. We can just call this name `pretrained` or `pretrained_model()`?

python/rikai/spark/sql/codegen/dummy.py
```diff
@@ -0,0 +1,60 @@
...
+
+from rikai.mixin import Pretrained
+from rikai.spark.sql.codegen.base import ModelSpec, Registry
+from rikai.spark.sql.exceptions import SpecError
+from rikai.spark.sql.model import NOURI_SPEC_SCHEMA
+
+__all__ = ["NoURIRegistry"]
+
+
+class DummyModelSpec(ModelSpec):
```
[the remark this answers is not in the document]
da-tubi: For now, every registry has a corresponding ModelSpec.

python/rikai/pytorch/models/fasterrcnn_resnet50_fpn.py
```diff
@@ -12,12 +12,28 @@
...
 from .torchvision import ObjectDetectionModelType
+from rikai.mixin import Pretrained
+from rikai.spark.sql.model import ModelSpec
+from rikai.spark.sql.codegen.dummy import DummyModelSpec
 
 __all__ = ["MODEL_TYPE"]
 
 
-class FasterRCNNModelType(ObjectDetectionModelType):
+class FasterRCNNModelType(ObjectDetectionModelType, Pretrained):
```
Renkai, line 25: In this class, `load_pretrained_model` and `load_model` both have a name in pattern `load_xx`, but the first one just returns a model to the caller, the second one will change the property of `self`. Can we design a name pattern that can help users to distinguish the kind of activities from the pattern of method names?

I propose we can name the first kind of activity `fetch_xx` that just returns a value and will not change `self`, `load_xx` are methods that will change the content of `self`.

da-tubi: see the latest commit

Renkai (approved):
LGTM

da-tubi:
https://github.com/eto-ai/rikai/pull/579/files#diff-5fd91b22b7c04a9dd0cce51026be4b084dacc8dd202a0a3f3a1eaf36e2ca8900

The `find_model` auxiliary method is not a good solution. We may improve it later.

Renkai:
I made an example of how we can move the template code into the mixin trait, not sure if we can do a similar thing in Python, paste it here for potential reference from the future.
```scala
object MixinExample {
  trait Model

  class DummyModel extends Model

  class OtherModel extends Model

  abstract class ModelType {
    def find_model(model: Model): Unit
  }

  trait Pretrained extends ModelType {

    def pretrained_model(): Model

    abstract override def find_model(model: Model): Unit = {
      if (model.isInstanceOf[DummyModel]) {
        super.find_model(pretrained_model())
      } else {
        super.find_model(model)
      }
    }
  }

  class AModelType extends ModelType {

    override def find_model(model: Model): Unit = {
      println("call AModelType find model")
    }
  }

  class PretrainedAModelType extends AModelType with Pretrained {
    override def pretrained_model(): Model = {
      println("call pre-trained model")
      new OtherModel
    }
  }

  def main(args: Array[String]): Unit = {
    val t = new PretrainedAModelType
    println("try call dummy model")
    t.find_model(new DummyModel)
    println("=" * 80)
    println("try call other model")
    t.find_model(new OtherModel)
  }
}

```

da-tubi:
Just tried the onnx version of ssd:
```
+    def pretrained_model(self) -> Any:
+        import onnx
+        from onnx_tf.backend import prepare
+        onnx_m = onnx.load("/tmp/ssd-12.onnx")
+        return prepare(onnx_m)
+
+    def load_model(self, spec: ModelSpec, **kwargs):
+        if isinstance(spec, DummyModelSpec):
+            self.model = self.pretrained_model()
+            self.spec = spec
+        else:
+            return super().load_model(spec, **kwargs)
```

```
+        results = self.model.run(images)
```
Using the tensorflow flavor, the input type does not match.
```
E                   TypeError: Value passed to parameter 'input' has DataType uint8 not in list of allowed values: float16, bfloat16, float32, float64, int32
```

PR #581, opened 2022-03-24
Support offset in Rikai parquet dataset

eddyxu:


python/rikai/parquet/dataset.py
```diff
@@ -61,18 +59,14 @@ class Dataset:
         a dataset URI, i.e., "s3://foo/bar"
     columns : List[str], optional
         To read only given columns
-    shuffle : bool, optional
```
eddyxu, line 64: Remove shuffler because we can use pytorch 1.8+ ShuffleDataLoader now

changhiskhan (approved):

PR #594, opened 2022-04-01
Assembly dependencies cleaning

### LINKED ISSUE #588: Assembly dependencies cleaning
## Remove cats and shapeless, we will reduce half of the assembly jar size
See https://github.com/eto-ai/rikai/pull/419

## Remove Scala Standard Library
image

## Remove antlr4 (use the Spark built-in one)
image

da-tubi:
closes #588 
Reduce the assembly jar size from 28.5M to 12M by:
+ Replace circe with Spark-shipped json4s
+ Remove scala-library
+ Remove enableIf.scala

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -79,7 +79,7 @@ libraryDependencies ++= {
 
   Seq(
     "org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided,
-    "com.thoughtworks.enableIf" %% "enableif" % enableifVersion exclude (
+    "com.thoughtworks.enableIf" %% "enableif" % enableifVersion % "compile-internal" exclude (
       "org.scala-lang", "scala-reflect"
     ),
     "org.apache.spark" %% "spark-sql" % sparkVersion.value % Provided,
@@ -144,7 +144,7 @@ Antlr4 / antlr4Version := {
   if ("3.1.2".equals(sparkVersion.value)) "4.8-1"
   else "4.8"
 }
-
+Antlr4 / antlr4RuntimeDependency := "org.antlr" % "antlr4-runtime" % (Antlr4 / antlr4Version).value % Provided
 enablePlugins(Antlr4Plugin)
 
 Compile / doc / scalacOptions ++= Seq(
@@ -155,13 +155,6 @@ Compile / doc / scalacOptions ++= Seq(
 assembly / assemblyJarName := s"${name.value}-assembly-${sparkVerStr.value}_${scalaBinaryVersion.value}-${version.value}.jar"
 // Excluding Scala library jars, see https://github.com/sbt/sbt-assembly/tree/v1.2.0#excluding-scala-library-jars
 assemblyPackageScala / assembleArtifact := false
-assembly / assemblyExcludedJars := {
-  val cp = (assembly / fullClasspath).value
-  cp filter { x =>
-    x.data.getName.contains("antlr4-runtime") ||
-      x.data.getName.contains("enableif")
-  }
-}
 
 publishLocal := {
   val ivyHome = ivyPaths.value.ivyHome.get.getCanonicalPath
--- a/build.sbt
+++ b/build.sbt
@@ -73,8 +73,6 @@ libraryDependencies ++= {
   val scalaLoggingVersion = "3.9.4"
   val snappyVersion = "1.1.8.4" // Support Apple Silicon
   val scalatestVersion = "3.2.0"
-  val hadoopVersion = "3.2.3"
-  val circeVersion = "0.12.3"
   val mlflowVersion = "1.21.0"
   val enableifVersion = "1.1.8"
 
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -154,6 +153,14 @@ Compile / doc / scalacOptions ++= Seq(
 )
 
 assembly / assemblyJarName := s"${name.value}-assembly-${sparkVerStr.value}_${scalaBinaryVersion.value}-${version.value}.jar"
+assemblyPackageScala / assembleArtifact := false
```
Renkai, line 156: Could you add some comment lines about what are these lines doing?

da-tubi: See https://github.com/sbt/sbt-assembly/tree/v1.2.0#excluding-scala-library-jars
It is documented in the sbt-assembly README. Let me add a comment about it

build.sbt
```diff
@@ -154,6 +153,15 @@ Compile / doc / scalacOptions ++= Seq(
 )
 
 assembly / assemblyJarName := s"${name.value}-assembly-${sparkVerStr.value}_${scalaBinaryVersion.value}-${version.value}.jar"
+// Excluding Scala library jars, see https://github.com/sbt/sbt-assembly/tree/v1.2.0#excluding-scala-library-jars
+assemblyPackageScala / assembleArtifact := false
+assembly / assemblyExcludedJars := {
+  val cp = (assembly / fullClasspath).value
+  cp filter { x =>
```
Renkai, line 160: Is it possible to change the assembly inventory by changing the configurations of a dependency rather than filter by name?
Can we try the configs in these two files?
https://github.com/sbt/librarymanagement/blob/a83f75c02e91afa114c77dae1b5a8f084913b1e4/core/src/main/scala/sbt/librarymanagement/LibraryManagementSyntax.scala#L28-L36

https://github.com/sbt/librarymanagement/blob/a83f75c02e91afa114c77dae1b5a8f084913b1e4/core/src/main/scala/sbt/librarymanagement/Configuration.scala#L46-L58

------
I was a user of `gradle`, in `gradle`, we can set a dependency to `compile`(which is actually `compile` and `runtime`), or `compileOnly`, if it's `compileOnly`, the assembly plugin will not include it.

Renkai: I'm not sure, but I guess `Compile` in `sbt` is `compileOnly` in `gradle`, and `Runtime` in `sbt` is `Compile` in `gradle`.

darcy-shen: Just mark `enableif` as Provided, it will not be assembled. But I think it is better to exclude it explicitly. Because I think `Provided` is semanticly wrong.

darcy-shen: And antlr4-runtime could be marked as Provided, because it is truely provided via Apache Spark.

darcy-shen: Well, `antlr4-runtime` is brought by the sbt plugin, the `antlr4-runtime` brought by spark deps is already `Provided`.

Since `antlr4-runtime` has zero dependency, I think it is ok to exclude it by classpath filter.

darcy-shen: Well, `antlr4-runtime` is brought by the sbt plugin, the `antlr4-runtime` brought by spark deps is already `Provided`.

Since `antlr4-runtime` has zero dependency, I think it is ok to exclude it by classpath filter.

Renkai: I don't think `Provided` is semantically wrong, from the definition [but indicates you expect the JDK or a container to provide the dependency at runtime](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope). Here in our current scenario, the runtime of `enableif` is project build time, it do provide the dependency.

Renkai: I don't think `Provided` is semantically wrong, from the definition [but indicates you expect the JDK or a container to provide the dependency at runtime](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope). Here in our current scenario, the runtime of `enableif` is project build time, it do provide the dependency.

Renkai: I have a further idea, maybe we can make `enableif` an sbt plugin, like put `addSbtPlugin("com.thoughtworks.enableIf" % "enableif" % "1.1.8")` into `plugins.sbt`, then we don't have this bothering. But I just tried it, it failed, maybe we can do some additional work to make it a plugin.

darcy-shen: See https://stackoverflow.com/questions/35477974/add-a-compile-time-only-sub-project-dependency-in-sbt/35486527

We could use CompileOnly

darcy-shen: I googled about it when introducing enableif.scala

I though it would increase the complexity of the sbt build def.

darcy-shen: `compile-internal` is the solution for enableif.scala

Renkai (approved):
It's the best `build.sbt` I've ever seen!

PR #596, opened 2022-04-01
Allow set parquet block size via option

eddyxu:
Use 

```python
df.write.format("rikai").option("rikai.block.size", 12345).save("dest")
```

to customize block size in the resulted parquet files.

### DIFF
--- a/src/main/scala/ai/eto/rikai/RikaiOptions.scala
+++ b/src/main/scala/ai/eto/rikai/RikaiOptions.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Rikai authors
+ * Copyright 2022 Rikai authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


eddyxu:
Closes #589

src/main/scala/ai/eto/rikai/RikaiOptions.scala
```diff
@@ -21,4 +21,10 @@ private[rikai] class RikaiOptions(parameters: Map[String, String]) {
   /** Base path for the feature dataset
     */
   val path: String = parameters.getOrElse("path", "")
+
+  val defaultBlockSize: Int = 32 * 1024 * 1024
+
+  /** Parquet block size. */
+  val blockSize: Int =
+    parameters.getOrElse("rikai.block.size", s"${blockSize}").toInt
```
changhiskhan, line 29: Should the format str be defaultBlockSize?

eddyxu: Ah, was made a mistake in the latest cleanup

changhiskhan: how come the tests passed tho?

changhiskhan (approved):
nit: maybe add a test so the typo wouldn't pass?

Issue #550, opened 2022-02-17
Visualization convenience


changhiskhan:
### Goal
we want to deliver an easy experience for developers to visualize images and annotations. One key criterion to enable that is to support users to display annotations like Box2d types on Image types without having to deal with lower level APIs like PIL or OpenCV

### Current gap
As demonstrated in https://github.com/eto-ai/rikai/blob/main/python/tests/types/test_vision.py#L261, currently drawing a Label is still inconvenient in 2 ways:
1. the user is required to specify the exact position, which requires the user to know the associated bounding box (or other annotation)
2. the user must wrap the label text in a Text constructor call

### A few examples:

Currently to draw labeled bounding boxes on images, you have to do the following:
`Image | bbox1 | bbox2 | Text("label1", (x1, y1)) | Text("label2", (x2, y2))`
This is inconvenient because (x1, y1), (x2, y2) requires both:
1. Getting some anchor position out of bbox1 and bbox2 (xmin, xmax, ymin, ymax)
2. Some trial and error to figure out what to add/sub from the anchor position so there's no overlap between the label and the bounding box.
And this is all just for one image and 2 boxes/labels.

### Desired behavior
`Image | (bbox1 + "label1") | (bbox2 + "label2")` or `Image | (bbox1, "label1") | (bbox2, "label2")` (or something along those veins) and rikai will automatically put "label1" on the say top-left of bbox1 and label2 on the top-left of bbox2.
And to customize the behavior, the user can do something like `rikai.options.viz.label_anchor_position = "bottom-right"` or `rikai.options.viz.label_position_func = {Box2d: lambda bbox: (bbox.xmin-10, bbox.ymin+10)}`

### Some thoughts on what to do
1. Allow users to express <drawable> + <label> association.
  - For classification we want a label to be associated with the image (i.e., no bounding bound or other annotations). 
  - For object detection, each image will have zero or more pairs of associated annotations+label pairs.
  - For segmentation, each image will have zero or more pairs of associated mask/polygon + label pairs. 
2. Each association has default behavior for automatically configuring (most important is the position) the labels
3. Allow the user to customize the configurations of the labels

changhiskhan:
Some references:
1. Viz interface defined here: https://github.com/eto-ai/rikai/blob/main/python/rikai/viz.py
2. Each Rikai type has rendering hooks like: https://github.com/eto-ai/rikai/blob/main/python/rikai/types/geometry.py#L257

Renkai:
[...]

PR #602, opened 2022-04-05
add mixin ComposableDrawable

Renkai:
related issue: https://github.com/eto-ai/rikai/issues/550

### DIFF
--- a/notebooks/Visualize.ipynb
+++ b/notebooks/Visualize.ipynb
@@ -180,9 +180,9 @@
    "source": [
     "boxes = [Box2d.from_top_left(*ann[\"bbox\"]) for ann in annotations]\n",
     "height, width = image_metadata[\"height\"], image_metadata[\"width\"]\n",
-    "segmentations = [Mask.from_polygon(ann[\"segmentation\"], height=height, width=width) \n",
-    "            if ann[\"iscrowd\"] == 0 \n",
-    "            else Mask.from_coco_rle(ann[\"segmentation\"][\"counts\"], height=height, width=width) \n",
+    "segmentations = [Mask.from_polygon(ann[\"segmentation\"], height=height, width=width)\n",
+    "            if ann[\"iscrowd\"] == 0\n",
+    "            else Mask.from_coco_rle(ann[\"segmentation\"][\"counts\"], height=height, width=width)\n",
     "            for ann in annotations]\n",
     "labels = [coco.loadCats(ids=ann[\"category_id\"])[0][\"name\"] for ann in annotations]"
    ]
@@ -218,6 +218,8 @@
     "from functools import reduce\n",
     "\n",
     "from rikai.viz import Style\n",
+    "\n",
+    "boxes[0] = boxes[0] + \"label\"\n",
     "\n",
     "box_style = Style(color=\"yellow\", width=3)\n",
     "\n",
--- a/python/rikai/mixin.py
+++ b/python/rikai/mixin.py
@@ -23,6 +23,7 @@ from typing import BinaryIO, Mapping, Optional, Union
 
 # Third Party
 import numpy as np
+
 # Rikai
 from rikai.internal.uri_utils import uri_equal
 from rikai.io import open_uri
@@ -97,7 +98,9 @@ class ComposableDrawable(Drawable):
         """return a tuple of the left top axis (xmin, ymin)"""
 
     @abstractmethod
-    def aligned(self, base_drawable: ComposableDrawable, algorithm: str = "left-top"):
+    def aligned(
+        self, base_drawable: ComposableDrawable, algorithm: str = "left-top"
+    ):
         """return a new instance that is aligned with the base drawable, currently it has only `left-top` align, will
         add more if we need them"""
 
@@ -108,8 +111,10 @@ class ComposedDrawable(Drawable):
     def align(self):
         if len(self.inner_drawables) <= 1:
             return
-        self.inner_drawables = self.inner_drawables[:1] + [inner.aligned(self.inner_drawables[0]) for inner in
-                                                           self.inner_drawables]
+        self.inner_drawables = self.inner_drawables[:1] + [
+            inner.aligned(self.inner_drawables[0])
+            for inner in self.inner_drawables
+        ]
 
     def _render(self, render: "rikai.viz.Renderer", **kwargs) -> None:
         self.align()
@@ -117,7 +122,9 @@ class ComposedDrawable(Drawable):
             inner._render(render, **kwargs)
 
     def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
-        self.inner_drawables = [inner.__matmul__(style) for inner in self.inner_drawables]
+        self.inner_drawables = [
+            inner.__matmul__(style) for inner in self.inner_drawables
+        ]
 
     def __init__(self):
         self.inner_drawables = []
@@ -144,7 +151,7 @@ class Asset(ABC):
     """
 
     def __init__(
-            self, data: Optional[bytes] = None, uri: Union[str, Path] = None
+        self, data: Optional[bytes] = None, uri: Union[str, Path] = None
     ) -> None:
         assert (data is None) ^ (uri is None)
         assert data is None or isinstance(
--- a/python/rikai/mixin.py
+++ b/python/rikai/mixin.py
@@ -101,12 +101,14 @@ class ComposableDrawable:
     def aligned(
         self, base_drawable: ComposableDrawable, algorithm: str = "left-top"
     ):
-        """return a new instance that is aligned with the base drawable, currently it has only `left-top` align, will
+        """return a new instance that is aligned with the base drawable,
+         currently it has only `left-top` align, will
         add more if we need them"""
 
 
 class ComposedDrawable(Drawable):
-    """A drawable that includes multi sub composable drawables, will align them in rendering"""
+    """A drawable that includes multi sub composable drawables,
+     will align them in rendering"""
 
     def align(self):
         if len(self.inner_drawables) <= 1:
--- a/python/rikai/mixin.py
+++ b/python/rikai/mixin.py
@@ -115,7 +115,7 @@ class ComposedDrawable(Drawable):
             return
         self.inner_drawables = self.inner_drawables[:1] + [
             inner.aligned(self.inner_drawables[0])
-            for inner in self.inner_drawables
+            for inner in self.inner_drawables[1:]
         ]
 
     def _render(self, render: "rikai.viz.Renderer", **kwargs) -> None:
--- a/python/rikai/mixin.py
+++ b/python/rikai/mixin.py
@@ -108,7 +108,7 @@ class ComposableDrawable:
 
 class ComposedDrawable(Drawable):
     """A drawable that includes multi sub composable drawables,
-     will align them in rendering"""
+    will align them in rendering"""
 
     def align(self):
         if len(self.inner_drawables) <= 1:
--- a/python/rikai/mixin.py
+++ b/python/rikai/mixin.py
@@ -23,7 +23,6 @@ from typing import BinaryIO, Mapping, Optional, Union
 
 # Third Party
 import numpy as np
-
 # Rikai
 from rikai.internal.uri_utils import uri_equal
 from rikai.io import open_uri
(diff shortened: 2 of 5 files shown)


python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
         return style(self)
 
 
+class ComposableDrawable:
+    """Mixin for a drawable that can be composed with others"""
+
+    @abstractmethod
+    def xy_min(self):
```
da-tubi, line 97: rename to left_top ? xy_min is not straghtforward.

notebooks/Visualize.ipynb
```diff
@@ -219,6 +219,8 @@
     "\n",
     "from rikai.viz import Style\n",
     "\n",
+    "boxes[0] = boxes[0] + \"label\"\n",
```
da-tubi, line 222: does it change the output?

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
...
+
+
+class ComposedDrawable(Drawable):
+    """A drawable that includes multi sub composable drawables,
+    will align them in rendering"""
+
+    def align(self):
+        if len(self.inner_drawables) <= 1:
+            return
+        self.inner_drawables = self.inner_drawables[:1] + [
```
da-tubi, line 116: please add unit tests when `len(self.inner_drawables) > 2`

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
...
+            inner.aligned(self.inner_drawables[0])
+            for inner in self.inner_drawables
```
da-tubi, line 118: it seems the first one is aligning with itself

python/tests/types/test_vision.py
```diff
@@ -264,6 +261,27 @@ def test_draw_texts():
     assert np.array_equal(pil_image.to_numpy(), expected)
 
 
+def test_draw_box_with_label():
+    data = np.random.randint(0, 255, size=(100, 100, 3), dtype=np.uint8)
+    img = Image.from_array(data)
+
+    box1 = Box2d(1, 2, 10, 12)
+    box2 = Box2d(20, 20, 40, 40)
+    draw_boxes = img | (box1 + "label1") | box2.with_label("label2")
```
da-tubi, line 270: this pr adds new operator `+` to express

align with the base drawable first and then append the layer to the base drawable

da-tubi: it is an important design and I think you should describe about it in the pull request desc part.

da-tubi: well, there is a detailed desc in #550

da-tubi: commented in #550, for Mask, the `+` operator is also needed.

python/rikai/mixin.py
Renkai: good idea

notebooks/Visualize.ipynb
Renkai: yes, we should see something different after it released.

python/rikai/mixin.py
Renkai: Yes, all others will align with the first one, the first one won't change

python/rikai/mixin.py
Renkai: In our current usage, the `len` would not `>2`, because it's only a box2d and a label, if we should add some use cases that have a larger drawable count, we'd better find out a concrete scenario.

python/tests/types/test_vision.py
Renkai: Currently,  `+` is only an alias of `.with_label`, and it's only a convenient feature between `box2d` and `label`. If we should expand it to more classes, maybe we should design a coherent semantic of syntax sugar for all classes.

python/rikai/mixin.py
da-tubi: it seems you changed two of them into left_top

da-tubi: and there are still two xy_min to be replaced by left_top

python/rikai/mixin.py
darcy-shen: why not skip the first?

darcy-shen: how about skipping the first one?

da-tubi: how about skipping the first one?

darcy-shen (approved):
LGTM now

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
...
+        self.align()
+        for inner in self.inner_drawables:
+            inner._render(render, **kwargs)
+
+    def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
+        self.inner_drawables = [
+            inner.__matmul__(style) for inner in self.inner_drawables
+        ]
+
+    def __init__(self):
```
eddyxu, line 131: Lets move __init__ to be the first function in the class

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
...
+            inner.aligned(self.inner_drawables[0])
+            for inner in self.inner_drawables[1:]
+        ]
+
+    def _render(self, render: "rikai.viz.Renderer", **kwargs) -> None:
+        self.align()
+        for inner in self.inner_drawables:
+            inner._render(render, **kwargs)
+
+    def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
```
eddyxu, line 126: Can we add some document about what the syntax looks like using this/

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
...
+         currently it has only `left-top` align, will
+        add more if we need them"""
+
+
+class ComposedDrawable(Drawable):
+    """A drawable that includes multi sub composable drawables,
+    will align them in rendering"""
+
+    def align(self):
+        if len(self.inner_drawables) <= 1:
```
eddyxu, line 114: `if not self.inner_drawables`

python/rikai/types/geometry.py
```diff
@@ -102,6 +111,33 @@ class Box2d(ToNumpy, Sequence, ToDict, Drawable):
...
+    ):
+        (new_xmin, new_ymin) = base_drawable.left_top()
+        return Box2d(
+            new_xmin,
+            new_ymin,
+            self.xmax - self.xmin + new_xmin,
+            self.ymax - self.ymin + new_ymin,
+        )
+
+    def with_label(
```
eddyxu, line 128: cal this take more text customization, i.e., `font size,  background color` and et.

python/rikai/types/geometry.py
```diff
@@ -102,6 +111,33 @@ class Box2d(ToNumpy, Sequence, ToDict, Drawable):
     >>> draw.rectangle(box, fill="green", width=2)
     """
 
+    def left_top(self):
+        return self.xmin, self.ymin
+
+    def aligned(
+        self, base_drawable: ComposableDrawable, algorithm: str = "left-top"
+    ):
+        (new_xmin, new_ymin) = base_drawable.left_top()
```
eddyxu, line 120: I am confused.  What usecase is it for a newly aligned bbox with other object (assuming box2d)?

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
         return style(self)
 
 
+class ComposableDrawable:
+    """Mixin for a drawable that can be composed with others"""
+
+    @abstractmethod
+    def left_top(self):
```
eddyxu, line 97: isnt this hard code the way the position? what if an appropriate label position would be bottom / center ?

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
...
+        """return a tuple of the left top axis (xmin, ymin)"""
+
+    @abstractmethod
+    def aligned(
+        self, base_drawable: ComposableDrawable, algorithm: str = "left-top"
```
eddyxu, line 102: `algoritm => position`?

eddyxu: Probably find a shorter name for `base_drawable` as well. `_drawable` has already indicated via the type.

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
...
+    ):
+        """return a new instance that is aligned with the base drawable,
```
eddyxu, line 104: Lets follow with the numpy docstring style here?

https://numpydoc.readthedocs.io/en/latest/format.html

python/rikai/mixin.py
eddyxu: I kinda dont think this `left_top` and `aligned` belong to the same interface.

`left_top / position` seems to belong to `base_drawable`, while `aligned` is the one that attached to the `base_drawable` right?

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
...
+
+    def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
+        self.inner_drawables = [
+            inner.__matmul__(style) for inner in self.inner_drawables
+        ]
+
+    def __init__(self):
+        self.inner_drawables = []
+
+    def add_drawable(self, drawable: ComposableDrawable):
```
eddyxu, line 134: lets use shorter name here?

``add`` might be sufficient?

eddyxu:
In general , i have some questions w/ the interface design. please take a look

eddyxu (changes requested):
Have some questions in API design.

changhiskhan (commented):
A valiant attempt. Thanks for taking a stab at this.
The main issue I have is that the abstractions seem really confusing.
I see why you made the `left_top` method, because you want to introspect the positions, but 1) it's very brittle (you need a lot more than just left_top), and 2) it really only makes sense for box2d.

I'm wondering if it makes more sense to interpret the anchor positions (e.g., "top-left") at `_render` time for the "base" where we know exactly what gets drawn. This is more flexible across shapes.

@eddyxu wydt

python/rikai/types/geometry.py
Renkai: If we add more parameters here, then we can't make it a synonym of `+`, maybe we should consider remove the symbol version of this method.

python/rikai/types/geometry.py
Renkai: It's for implementing the `ComposableDrawable` trait, doable but not useful yet, maybe it would be eventually not useful, I would consider keeping it unimplemented until we really need it.

python/rikai/mixin.py
Renkai: In my idea, `aligned` would apply some algorithm, `left_top` is an anchor that an algorithm may use. Because `left-top` is currently our only algorithm, so I only added this anchor, we may add other anchors later. I'm not sure if the anchor method would be useful in other places, so I did not add it in the base class.

python/rikai/mixin.py
```diff
@@ -90,6 +90,51 @@ def __matmul__(self, style: Union[dict, "rikai.viz.Style"]) -> Drawable:
...
+    @abstractmethod
+    def aligned(
+        self, base_drawable: ComposableDrawable, algorithm: str = "left-top"
+    ):
+        """return a new instance that is aligned with the base drawable,
+         currently it has only `left-top` align, will
+        add more if we need them"""
+
+
+class ComposedDrawable(Drawable):
```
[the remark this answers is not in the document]
Renkai: If we need a more general name for an anchor position (for 1d, 2d, 3d, 4d...), my limited Math and English knowledge can't give a reasonable one😅, do you have any ideas?

python/rikai/types/geometry.py
```diff
@@ -102,6 +111,33 @@ class Box2d(ToNumpy, Sequence, ToDict, Drawable):
...
+        )
+
+    def with_label(
+        self, text: str, color: str = get_option(CONF_RIKAI_VIZ_COLOR)
+    ) -> Drawable:
+        """return a ComposedDrawable with label on the box"""
+
+        composed = ComposedDrawable()
+        composed.add_drawable(self)
+        composed.add_drawable(Text(text, (self.xmin, self.ymin), color))
```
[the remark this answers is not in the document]
Renkai: Yes, maybe we can also add a `center` if you think the current architecture is ok.

python/rikai/mixin.py
darcy-shen: How about use the python function as the parameter.

In this case, `left_top` is a member of ComposableDrawable. Using python function as the parameter, we could separate left_top and the Mixin.

python/rikai/mixin.py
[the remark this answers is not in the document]
changhiskhan: well, i don't think my english skills are good enough to come up with 4d-space anchor positions either :)

But this is more than just about the naming of anchor positions.

darcy-shen:
> The main issue I have is that the abstractions seem really confusing.

I think so. I think the only interface introduced here is `Box2d + "label_text"`. The impl abstractions are not ideal, but it works. That's why I approved this pull request.

Renkai:
Replaced by a simplified version here https://github.com/eto-ai/rikai/pull/632

PR #617, opened 2022-04-08
Make label / candidates list to be one metadata of model type

### LINKED ISSUE #592: Make label / candidates list to be one metadata of model type
# Problem

Our model type only converts the tensors of label ids into `int` in the ModelType output, which makes the outputs not human readable, unless users convert id to text later by themselves.

# Solution

Model type can specify a callable or an object that returns the string/text representation of label ID  from detections. It also helps users who take off-the-shelf models and train them on new datasets with different labels.

changhiskhan:
Closes #592

### DIFF
--- a/python/rikai/pytorch/models/resnet.py
+++ b/python/rikai/pytorch/models/resnet.py
@@ -16,27 +16,32 @@
 ResNet: Deep Residual Learning for Image Recognition
 https://arxiv.org/pdf/1512.03385.pdf
 """
-
 import torchvision
 
-from rikai.pytorch.models.torch import ClassificationModelType
+from rikai.pytorch.models.torch import ClassificationModelType, classification_id_to_label_fn
 
 resnet18 = ClassificationModelType(
-    name="resnet18", pretrained_fn=torchvision.models.resnet18
+    name="resnet18", pretrained_fn=torchvision.models.resnet18,
+    id_to_label_fn=classification_id_to_label_fn
 )
 resnet34 = ClassificationModelType(
-    name="resnet34", pretrained_fn=torchvision.models.resnet34
+    name="resnet34", pretrained_fn=torchvision.models.resnet34,
+    id_to_label_fn=classification_id_to_label_fn
 )
 resnet50 = ClassificationModelType(
-    name="resnet50", pretrained_fn=torchvision.models.resnet50
+    name="resnet50", pretrained_fn=torchvision.models.resnet50,
+    id_to_label_fn=classification_id_to_label_fn
 )
 resnet101 = ClassificationModelType(
-    name="resnet101", pretrained_fn=torchvision.models.resnet101
+    name="resnet101", pretrained_fn=torchvision.models.resnet101,
+    id_to_label_fn=classification_id_to_label_fn
 )
 resnet152 = ClassificationModelType(
-    name="resnet152", pretrained_fn=torchvision.models.resnet152
+    name="resnet152", pretrained_fn=torchvision.models.resnet152,
+    id_to_label_fn=classification_id_to_label_fn
 )
 # Make default resnet to be ResNet-50
 resnet = ClassificationModelType(
-    name="resnet", pretrained_fn=torchvision.models.resnet50
+    name="resnet", pretrained_fn=torchvision.models.resnet50,
+    id_to_label_fn=classification_id_to_label_fn
 )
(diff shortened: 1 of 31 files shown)


python/rikai/pytorch/models/resnet.py
```diff
@@ -16,27 +16,41 @@
...
+from rikai.pytorch.models.torch import (
+    classification_id_to_label_fn,
+    ClassificationModelType,
+)
 
 resnet18 = ClassificationModelType(
-    name="resnet18", pretrained_fn=torchvision.models.resnet18
+    name="resnet18",
+    pretrained_fn=torchvision.models.resnet18,
+    id_to_label_fn=classification_id_to_label_fn,
```
eddyxu, line 29: can we provide a default value for `id_to_label_fn`?
Seems all classification models share the same one, similar situation applies to `object detections` as well.

python/rikai/pytorch/models/torch.py
```diff
@@ -18,6 +18,7 @@
 from abc import ABC
 from typing import Any, Callable, Optional
 
+import requests
```
eddyxu, line 21: why `requests` is needed here

python/rikai/spark/sql/model.py
```diff
@@ -155,6 +156,19 @@ def model_type(self) -> "ModelType":
     def load_model(self) -> Any:
         """Load the model artifact specified in this spec"""
 
+    def load_id_to_label_fn(self) -> Optional[Callable]:
```
eddyxu, line 159: shall we use `label_fn` for short?  we can make the typing to reflect the input/output types, i.e.

```Callable[[int], str]```.

File: python/rikai/pytorch/models/resnet.py (hunk 1 above)
Comment (changhiskhan) [reply python/rikai/pytorch/models/resnet.py:29]: yeah, this PR have default ones setup for classification, object detection, and key point detection

File: python/rikai/pytorch/models/torch.py (hunk 2 above)
Comment (changhiskhan) [reply python/rikai/pytorch/models/torch.py:21]: per pytorch docs, it's fetching the default classification model labels from github

File: python/rikai/spark/sql/model.py (hunk 3 above)
Comment (changhiskhan) [reply python/rikai/spark/sql/model.py:159]: yeah, that's reasonable. I'll make that change

File: python/rikai/pytorch/models/resnet.py (hunk 1 above)
Comment (eddyxu) [reply python/rikai/pytorch/models/resnet.py:29]: can we make that func to be a default value in `ClassificationModelType`'s constructor, so that we do need to specify it here in the most of the case.

The reason that we need `pretrained_fn` there was that each model have different method to load pretrained models.

File: python/rikai/pytorch/models/torch.py (hunk 2 above)
Comment (eddyxu) [reply python/rikai/pytorch/models/torch.py:21]: ok, I see now. it might be fine for `requests`. Trying to see whether we can limit the exposure of these non-torch / ml essential dependencies. 

`requests` might be just too widely used to hide at this point.

Comment (changhiskhan) [reply python/rikai/pytorch/models/torch.py:21]: yeah I wouldn't be overly concerned about `requests` being included here

**Comment by darcy-shen** [comment]
Could you provide a better pull request title?

File: python/rikai/pytorch/models/ssd_class_scores.py [hunk 4]
```diff
@@ -20,7 +20,10 @@
 from torchvision.models.detection.ssd import SSD
 from torchvision.ops.boxes import batched_nms, clip_boxes_to_image
 
-from rikai.pytorch.models.torch import ObjectDetectionModelType
+from rikai.pytorch.models.torch import (
+    detection_label_fn,
```
eddyxu, line 24: dont need to import this function?

python/rikai/spark/sql/model.py
```diff
@@ -155,6 +156,19 @@ def model_type(self) -> "ModelType":
     def load_model(self) -> Any:
         """Load the model artifact specified in this spec"""
 
+    def load_label_fn(self) -> Optional[Callable]:
+        """Load the function that maps label id to human-readable string"""
```
eddyxu, line 160: should we also add support from SQL parser as well?

eddyxu (approved):
Some small nits. 

Lets file a ticket to add support of this function from SQL?

python/rikai/pytorch/models/resnet.py
changhiskhan: fixed

python/rikai/spark/sql/model.py
changhiskhan: made #626

PR #619, opened 2022-04-08
Model types for Sklearn

eddyxu:


python/rikai/sklearn/__ini__.py
```diff
@@ -0,0 +1,13 @@
+#  Copyright 2022 Rikai Authors
```
darcy-shen, line 1: file name missing t

darcy-shen (approved):

python/tests/spark/sql/codegen/test_sklearn_flavor.py
```diff
@@ -88,7 +86,7 @@ def test_sklearn_random_forest(mlflow_tracking_uri: str, spark: SparkSession):
         rikai.mlflow.sklearn.log_model(
             model,
             artifact_path="model",
-            schema="long",
+            model_type="random_forest_classification",
```
changhiskhan, line 89: as a user, how would they know to input this string as the model type?

eddyxu: Via document https://github.com/eto-ai/rikai/issues/607

PR #628, opened 2022-04-13
Change height and width order in parameters

eddyxu:


darcy-shen:
Other in-consist order is in https://github.com/eto-ai/rikai/tree/main/contrib/coco

darcy-shen (approved):
LGTM

PR #632, opened 2022-04-17
Box2d label

Renkai:
I also tried to implement something like this
```
    sugar_boxes = (
        img
        | box1.with_label('label1') @ {"color": "green", "width": 10}
        | box2.with_label('label2') @ {"color": "green", "width": 10}
    )
```

But it seems we can't do it by just returning a list for `with_label`. If we need to enable this sugar, maybe we still need something like `ComposedDrawable`.

### DIFF
--- a/python/rikai/types/geometry.py
+++ b/python/rikai/types/geometry.py
@@ -23,6 +23,7 @@ from typing import List, Optional, Sequence, Tuple, Union
 
 import numpy as np
 from PIL import Image, ImageDraw
+from rikai.conf import CONF_RIKAI_VIZ_COLOR, get_option
 
 from rikai.mixin import Drawable, ToDict, ToNumpy
 from rikai.spark.types.geometry import (
@@ -34,6 +35,8 @@ from rikai.spark.types.geometry import (
 from rikai.types import rle
 
 __all__ = ["Point", "Box3d", "Box2d", "Mask"]
+
+from rikai.viz import Text
 
 
 class Point(ToNumpy, ToDict):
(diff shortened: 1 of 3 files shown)


python/rikai/types/geometry.py
```diff
@@ -392,6 +395,25 @@ def iou(
             return iou_arr[0]
         return iou_arr
 
+    def with_label(
+        self, text: str, color: str = get_option(CONF_RIKAI_VIZ_COLOR)
```
eddyxu, line 399: Nits:  usually python will take `color: Option[str] = None`, and within the function itself, try to check 

```
if color is None:
    color = get_option(...)
```

python/rikai/types/geometry.py
```diff
@@ -392,6 +395,25 @@ def iou(
...
+    ):
+        """Spawn a list with current box and a `python.rikai.viz.Text`
```
eddyxu, line 401: Lets rephrase the summary of a function to what is doing, instead of what it returns?

eddyxu (commented):
Overall lgtm. small nits

python/rikai/types/geometry.py
Renkai: image
I see the current pattern many times in the master codebase, should we change them all? I chose the current way that's because I want to be consistent with existing code base.

python/rikai/types/geometry.py
```diff
@@ -392,6 +395,26 @@ def iou(
...
+            The text content of that label
+        color: str
+            The color of the text, will have a default value if not given.
+
+        Returns
+        -------
+        list
+            [self, the_label]
+        """
+        return [self, Text(text, (self.xmin, self.ymin), color)]
```
darcy-shen, line 416: the label and the text are overlapped, is it the expected behavior

python/rikai/types/geometry.py
darcy-shen: I agree with Eddy, and we should change them all. But I think it should be done in a separate pull request.

python/rikai/types/geometry.py
darcy-shen: I think for now, a hard-coded shift (shift_x, shift_y) is sufficient.

darcy-shen (changes requested):
A hard-coded or font-size aware shift is expected for with_label

image

python/rikai/types/geometry.py
changhiskhan: yeah, for now just have a hard-coded padding/margin. Later on we can add additional options as needed

Renkai:
use https://github.com/eto-ai/rikai/pull/650 instead

PR #635, opened 2022-04-19
Support partitionBy with Rikai dataframe writer in Spark

eddyxu:
`df.write.rikai()` supports `partitionBy(col, ...)`

### DIFF
--- a/src/main/scala/ai/eto/rikai/RikaiOptions.scala
+++ b/src/main/scala/ai/eto/rikai/RikaiOptions.scala
@@ -33,6 +33,19 @@ private[rikai] class RikaiOptions(parameters: Map[String, String]) {
     parameters
       .filterKeys(k => !RikaiOptions.excludedKeys(k))
       .toMap
+
+  /** Columns specified via df.partitionBy() */
+  val partitionColumns: Option[Seq[String]] =
+    parameters.get("__partition_columns") match {
+      case Some(cols) =>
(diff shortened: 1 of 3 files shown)


eddyxu:
Closes #634

src/main/scala/ai/eto/rikai/RikaiOptions.scala
```diff
@@ -33,6 +33,19 @@ private[rikai] class RikaiOptions(parameters: Map[String, String]) {
...
       .filterKeys(k => !RikaiOptions.excludedKeys(k))
       .toMap
+
+  /** Columns specified via df.partitionBy() */
+  val partitionColumns: Option[Seq[String]] =
+    parameters.get("__partition_columns") match {
+      case Some(cols) =>
+        Some(
+          cols
+            .substring(1, cols.length - 1)
```
changhiskhan, line 43: why do we need this `subtring` here?

eddyxu: The value in `parameters: Map[string, string]` is something like

```
Map(__partition_columns -> ["label"], path -> /var/folders/m1/vyb5yj4n5cb17gv371g09k280000gn/T/rikai6565224676268032985/dataset)
```

the string value includes double quote

changhiskhan (approved):

PR #648, opened 2022-04-29
scala UDF image to construct a Image UDF from a URI

### LINKED ISSUE #645: `to_image` or `image` automatically registered Scala UDF
I will setup a Rikai production airflow dag, and I found it frustrating that we have to register a python UDF and then use it.

I want to setup the airflow dag using only rikai sql, I do not want to make a python whl (just to register the python UDF) for this production dag.

I think at least `to_image` should be re-impl in Scala, or we just use another udf name, like `image`.

da-tubi:
close #645

### DIFF
--- a/src/main/scala/ai/eto/rikai/sql/spark/RikaiSparkSessionExtensions.scala
+++ b/src/main/scala/ai/eto/rikai/sql/spark/RikaiSparkSessionExtensions.scala
@@ -33,7 +33,7 @@ import org.apache.spark.sql.catalyst.expressions.{
 }
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.rules.Rule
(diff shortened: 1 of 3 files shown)


changhiskhan:
test failure

build.sbt
```diff
@@ -137,6 +137,22 @@ scalacOptions ++= {
 
 Test / parallelExecution := false
 Test / fork := true
+import Tests._
+{
+  def groupByFirst(tests: Seq[TestDefinition]) =
```
changhiskhan, line 142: Not sure I understand the problem. How did your change touch the Registry?

changhiskhan (commented):
I'm concerned about hacking the testing like this.
Can we chat on slack when you're up?

build.sbt
darcy-shen: Here is the official documentation of SBT: https://www.scala-sbt.org/1.x/docs/Testing.html#Forking+tests

changhiskhan (approved):

PR #650, opened 2022-05-03
Box2d label with `Draw`

Renkai:
Similar to https://github.com/eto-ai/rikai/pull/632, but extend `Draw` instead of using a list.
It can have a sugar
```
.... @ {"color": "green", "width": 10}
```
which list hasn't.

part of https://github.com/eto-ai/rikai/issues/550

### DIFF
--- a/python/rikai/types/geometry.py
+++ b/python/rikai/types/geometry.py
@@ -416,7 +416,16 @@ class Box2d(ToNumpy, Sequence, ToDict, Drawable):
         """
         # TODO add font size
         return Draw(
-            [self, Text(text, (int(self.xmin), int(self.ymin)), color)]
+            [
+                self,
+                Text(
+                    # `ymin - 10` to let the label don't overlap
+                    # with box by default find a better way later
+                    text,
+                    (int(self.xmin), int(min(self.ymin - 10, 0))),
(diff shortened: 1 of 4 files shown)


Renkai:
I want to set up the font size for text labels, but I found it complex, made it an independent issue.
https://github.com/eto-ai/rikai/issues/655

python/rikai/types/geometry.py
```diff
@@ -392,6 +394,31 @@ def iou(
...
+            The color of the text,
+            will have a default value if not given.
+        Returns
+        -------
+        box_with_label
+            rikai.viz.Draw
+        """
+        # TODO add font size
+        return Draw(
+            [self, Text(text, (int(self.xmin), int(self.ymin)), color)]
```
darcy-shen, line 419: as I suggested in #632 , we should shift the (xmin, ymin) like (xmin, ymin-10), and 10 should be the default height (or a bit bigger) of the font size

changhiskhan (commented):
let's automatically add some margin by default so the label doesn't overlap with the bounding box. other than that this looks a lot simpler. nice work.

darcy-shen (approved):
image

PR #653, opened 2022-05-04
DRAFT: Add the ability to write a pandas dataframe to disk in Rikai format without needing a live spark session

changhiskhan:
Open issues:
1. I had to change how Box2d is serialized/deserialized to make it work (otherwise the elements are written out of order)
2. This probably means that other shapes are like also wrong right now, but I wanted to solicit feedback before testing/changing those
3. Right now you still need to give it an explicit StructType schema so the resulting dataset is readable by spark and it knows when to convert the Rikai types into native types. It would be nice to add an option to just infer the types based on a sample of rows in the DataFrame.

### DIFF
--- a/python/rikai/parquet/writer.py
+++ b/python/rikai/parquet/writer.py
@@ -16,17 +16,26 @@
 """
 
 import json
+from typing import Union
+
+import pandas as pd
 import pyarrow as pa
+import pyarrow.dataset as ds
 import pyarrow.parquet as pq
+
 from rikai.mixin import ToDict
 from rikai.parquet.dataset import Dataset as RikaiDataset
-import pandas as pd
 
 __all__ = ["df_to_rikai"]
 
 
 def df_to_rikai(
-    df: pd.DataFrame, dest_path: str, schema: "pyspark.sql.types.StructType"
+    df: pd.DataFrame,
+    dest_path: str,
+    schema: "pyspark.sql.types.StructType",
+    partition_cols: Union[str, list] = None,
+    max_rows_per_file: int = None,
+    mode: str = None,
 ):
     """Write the given pandas dataframe to the given location using the
     specified (pyspark) schema
@@ -40,6 +49,15 @@ def df_to_rikai(
     schema: StructType
         The pyspark schema required to convert Rikai types and make the
         resulting rikai dataset readable by spark
+    partition_cols: str or list, default None
+        Columns to partition on
+    max_rows_per_file: int, default None
+        passed to Arrow Dataset.write_dataset
+    mode: str, default None
+        This controls `existing_data_behavior` in pyarrow.
+        {'error', 'overwrite_or_ignore', 'delete_matching'}
+        None will default to Pyarrow's current default ('error')
+        https://arrow.apache.org/docs/python/generated/pyarrow.dataset.write_dataset.html
     """
     schema_json_str = schema.json()
     converted_df = conv(df, json.loads(schema_json_str))  # convert Rikai types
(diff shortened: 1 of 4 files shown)


python/rikai/parquet/writer.py
```diff
@@ -0,0 +1,125 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""Tools to write Rikai format datasets to disk
+"""
+
+import json
+import pyarrow as pa
+import pyarrow.parquet as pq
+from rikai.mixin import ToDict
```
eddyxu, line 21: isort

python/rikai/parquet/writer.py
```diff
@@ -0,0 +1,125 @@
...
+from rikai.parquet.dataset import Dataset as RikaiDataset
+import pandas as pd
+
+__all__ = ["df_to_rikai"]
+
+
+def df_to_rikai(
+    df: pd.DataFrame, dest_path: str, schema: "pyspark.sql.types.StructType"
```
eddyxu, line 29: wonder whether we can either infer the scheme, or use a framework agnostic schema ?

python/rikai/parquet/writer.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _conv_udt(v, typ):
+    """Convert the UDT value to a serializable format"""
+    if isinstance(v, ToDict):
+        return v.to_dict()
+    else:
+        assert typ["type"] == "udt"
+        udt = RikaiDataset._find_udt(typ["pyClass"])
+        return udt.serialize(v)
```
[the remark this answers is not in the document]
eddyxu: we can probably to make a new interface to make the rikai UDT type directly returns a dict, instead of using spark's UDT interface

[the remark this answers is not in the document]
changhiskhan: > So this one still needs pyspark dependency (returns `pyspark.sql.Row`, just not live spark)?

yeah, and also the schema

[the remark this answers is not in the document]
changhiskhan: > we can probably to make a new interface to make the rikai UDT type directly returns a dict, instead of using spark's UDT interface

yeah right now what we have isn't ideal. i'm using the ToDict mixin for now.

python/rikai/parquet/writer.py
changhiskhan: there are two issues here. I think inferring the schema by default is doable (just sample a few rows and check).
spark row metadata has a lot going on so it'd be somewhat difficult to make it framework agnostic?

eddyxu: we do have the code in ModelType.schema to define a schema in `string` tho. Can we extend that?

> I think inferring the schema by default is doable (just sample a few rows and check).

Our previous experience was very troublesome on schema inference tho. cc @bobingm.

python/rikai/parquet/writer.py
```diff
@@ -0,0 +1,125 @@
...
+        The pyspark schema required to convert Rikai types and make the
+        resulting rikai dataset readable by spark
+    """
+    schema_json_str = schema.json()
+    converted_df = conv(df, json.loads(schema_json_str))  # convert Rikai types
+    # The spark schema is required so Spark can read it successfully
+    table = pa.Table.from_pandas(converted_df).replace_schema_metadata(
+        {RikaiDataset.SPARK_PARQUET_ROW_METADATA: schema_json_str}
+    )
+    pq.write_to_dataset(table, dest_path, use_legacy_dataset=False)
```
eddyxu, line 50: lets also support partitionBy?

Does this overwrite existing data by default? How to control the number of files and etc?

python/rikai/parquet/writer.py
changhiskhan: would love to learn what problems you ran into with schema inferencing (specifically for this use case i mean)

eddyxu (approved):

PR #657, opened 2022-05-07
DRAFT: pandas and arrow extensions for type inference

changhiskhan:
Pandas/Arrow extensions may make it possible to
easily read/write the Rikai types to/from parquet datasets.
For usage checkout the new `tests/types/test_pandas.py` file.
For a quick note on the proposed design, see top of `rikai/types/pandas.py`

Two caveats:
1. Rikai types (e.g., Image) must be a valid storage type. Here I've
   made it subclass dict. This clearly needs more thought on how to
   actually represent the Rikai types.
2. The current design does not work for nested types. For nested types,
   if we find that most ingested datasets have a similar structure,
   we *could* implement extension types that read/write custom StructArray,
   though I'm skeptical this is flexible enough. The other option is
   to fallback on getting type information from pandas dataframe columns
   where the dtype is object and elements are list-like or dict. The
   trade-off here is performance vs accuracy of course.

### DIFF
--- a/python/rikai/types/geometry.py
+++ b/python/rikai/types/geometry.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 from enum import Enum
 from numbers import Real
+from types import MappingProxyType
 from typing import List, Optional, Sequence, Tuple, Union
 
 import numpy as np
@@ -75,7 +76,7 @@ class Point(ToNumpy, ToDict):
         return {"x": self.x, "y": self.y, "z": self.z}
 
 
-class Box2d(ToNumpy, Sequence, ToDict, Drawable):
+class Box2d(ToNumpy, Sequence, ToDict, Drawable, dict):
     """2-D Bounding Box, defined by ``(xmin, ymin, xmax, ymax)``
 
     Attributes
@@ -115,6 +116,12 @@ class Box2d(ToNumpy, Sequence, ToDict, Drawable):
         self.ymin = float(ymin)
         self.xmax = float(xmax)
         self.ymax = float(ymax)
+        dict.__init__(self, {
+            'xmin': self.xmin,
+            'ymin': self.ymin,
+            'xmax': self.xmax,
+            'ymax': self.ymax
+        })
 
     @classmethod
     def from_center(
--- a/python/rikai/types/pandas.py
+++ b/python/rikai/types/pandas.py
@@ -1,0 +1,197 @@
+#  Copyright 2022 Rikai Authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+"""Pandas extensions
+
+For each Rikai type, we need the following:
+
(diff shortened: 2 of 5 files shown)


python/rikai/types/geometry.py
```diff
@@ -115,6 +116,12 @@ def __init__(self, xmin: float, ymin: float, xmax: float, ymax: float):
         self.ymin = float(ymin)
         self.xmax = float(xmax)
         self.ymax = float(ymax)
+        dict.__init__(self, {
```
eddyxu, line 119: does this store the values 2 times?

eddyxu: Wondering whether a box is both an array and dict breaks sth?
like `box[0] == box["xmin"] == box.xmin` .

python/tests/types/test_pandas.py
```diff
@@ -0,0 +1,55 @@
...
+
+    df_rt = table.to_pandas()
+    assert len(df_rt) == len(s)
+    assert df_rt.c.dtype == 'image'
+
+    # Check roundtrip pandas<>parquet
+    pq.write_to_dataset(table, str(tmp_path))
+    df_from_parquet = pd.read_parquet(str(tmp_path))
+    assert len(df_from_parquet) == len(s)
+    assert df_from_parquet.c.dtype == 'image'
```
eddyxu, line 54: this feels magical 💯

python/tests/types/test_pandas.py
```diff
@@ -0,0 +1,55 @@
...
+    ])
+    assert arrow_type.to_pandas_dtype().name == 'image'
+
+    df_rt = table.to_pandas()
+    assert len(df_rt) == len(s)
+    assert df_rt.c.dtype == 'image'
+
+    # Check roundtrip pandas<>parquet
+    pq.write_to_dataset(table, str(tmp_path))
+    df_from_parquet = pd.read_parquet(str(tmp_path))
```
eddyxu, line 52: This format is not inter-operable with Spark, right? Is the plan to write the spark schema metadata as a following PR?

Can `pd.read_parqeut()` read out a dataset generated from `spark.write.format("rikai")` ?

python/rikai/types/geometry.py
changhiskhan: Oh for Box2d this was some leftover hack. No need for Box2d to subclass both.

python/tests/types/test_pandas.py
changhiskhan: Right unfortunately this is not compatible with Spark anymore as it stands. What we would have to do (I think) is:
1. Create a custom pandas accessor (eg “df.rikai.save(…)” that takes care of writing Spark metadata
2. Customize the Spark Rikai writer to also add the equivalent pandas/arrow metadata.

python/rikai/types/pandas.py
```diff
@@ -0,0 +1,394 @@
...
+        dtype = image_arrow_type
+        storage = pa.array(self.data, type=dtype.storage_type)
+        return pa.ExtensionArray.from_storage(dtype, storage)
+
+
+dict_to_box = np.vectorize(lambda d: T.Box2d(**d))
+
+
+@register_dataframe_accessor('rikai')
+class RikaiDataFrameAccessor:
```
changhiskhan, line 295: This is the new addition for nested types @eddyxu

python/tests/types/test_pandas.py
```diff
@@ -0,0 +1,88 @@
...
+    assert len(df_from_parquet) == len(s)
+    assert df_from_parquet.c.dtype == 'image'
+    assert isinstance(df_from_parquet.c[0], Image)
+
+
+def test_nested(tmp_path):
```
changhiskhan, line 58: Here’s the test case. I need to dig a little more into arrow internals to see if I can get native support for this kind of nesting to work properly.

PR #658, opened 2022-05-09
[no ci] Separate RegistryTest out

da-tubi:
#648 is meant to deliver the UDF `image` without having to register `to_image` in Python.

This pull request is part of #648 but does not belong to #648 .

On the main branch, we could use `make test` to reproduce https://github.com/eto-ai/rikai/runs/6300123281?check_suite_focus=true 

It happens occasionally.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -137,6 +137,8 @@ scalacOptions ++= {
 
 Test / parallelExecution := false
 Test / fork := true
+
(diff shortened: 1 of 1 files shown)


changhiskhan (approved):
can you also add the sbt docs link (https://www.scala-sbt.org/1.x/docs/Testing.html#Forking+tests) in the code comments here as well?

PR #659, opened 2022-05-09
[POC] Experimental Postgres support

eddyxu:
A proof of concept work to run Rikai in PostgreSQL.

### DIFF
--- a/experimental/pg/rikai/experimental/pg/model.py
+++ b/experimental/pg/rikai/experimental/pg/model.py
@@ -30,22 +30,30 @@ class PgModel:
         return f"PgModel({self.model})"
 
     def predict(self, img):
-        tensor = torch.tensor(self.model.transform()(Image(img["uri"]).to_numpy()))
+        tensor = torch.tensor(
+            self.model.transform()(Image(img["uri"]).to_numpy())
+        )
         preds = self.model.predict([tensor])[0]
 
-        return [{
-            "label": pred["label"],
-            "label_id": pred["label_id"],
-            "score": pred["score"],
-            "box": ((pred["box"].xmin, pred["box"].ymin), (pred["box"].xmax, pred["box"].ymax))
-        } for pred in preds]
+        return [
+            {
+                "label": pred["label"],
+                "label_id": pred["label_id"],
+                "score": pred["score"],
+                "box": (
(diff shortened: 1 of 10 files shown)


experimental/pg/rikai/experimental/pg/model.py
```diff
@@ -0,0 +1,72 @@
...
+    # TODO: move load model into rikai core.
+    conf = {
+        "version": "1.0",
+        "name": f"load_{model_type}",
+        "flavor": flavor,
+        "modelType": model_type,
+        "uri": uri,
+    }
+    if uri:
+        spec = FileModelSpec(conf)
```
changhiskhan, line 67: should models just be stored in some pg metadata tables?

experimental/pg/sql/rikai--0.1.sql
```diff
@@ -0,0 +1,61 @@
+-- Create Schema to contain all Rikai functionality
+CREATE SCHEMA IF NOT EXISTS ml;
+
+-- Semantic Types
+CREATE TYPE image AS (uri TEXT);
+
+CREATE TYPE detection AS (label TEXT, label_id int, box box, score real);
```
changhiskhan, line 7: where's the box type?

experimental/pg/rikai/experimental/pg/model.py
changhiskhan: oh hmmm actually if the model inf/train needs to be externalized then maybe not

experimental/pg/sql/rikai--0.1.sql
eddyxu: Box and Polygon `Mask in rikai` are already supported. https://www.postgresql.org/docs/current/datatype-geometric.html

experimental/pg/rikai/experimental/pg/model.py
eddyxu: model specs are stored in `ml.models` table. It acts the same as in Spark's `CREATE MODEL /  SHOW MODELS`. 
So in this implementation, it treats `ml` schema / namespace as the ML catalog in Spark

eddyxu:
Move the code to a separate repo https://github.com/eto-ai/rikai-pg,  giving its early maturity.

Issue #640, opened 2022-04-23
`df.saveAsTable()` does not take partitionBy


eddyxu:
```
df.write.partitionBy("col").format("rikai").saveAsTable("blabla")
```
does not write files in partitions.

But  the following code, without specifying `rikai` format does write in partitions.

```
df.write
      .partitionBy("col")
      .mode("overwrite")

      .saveAsTable("parts")
```

Renkai:
I would take a look

ref:
https://github.com/eto-ai/rikai/commit/70e1ccb9d5c22bfc07aaa4d23148f42ad58cccb5

Renkai:
What's the environment for reproducing? local cluster, an open-source cluster of databricks platform?

Renkai:
Save to path method works fine, try to find the special part of saveAsTable.
https://github.com/eto-ai/rikai/blob/70e1ccb9d5c22bfc07aaa4d23148f42ad58cccb5/src/test/scala/ai/eto/rikai/RikaiRelationTest.scala#L61

Renkai:
By my local test 
```
df.write.format("rikai").partitionBy("col").saveAsTable("blabla")
```
also fails, so it's not a config overwritten accident. Maybe that's because the `rikai` format can't satisfy some condition for `partitionBy`

Renkai:
https://github.com/eto-ai/rikai/blob/70e1ccb9d5c22bfc07aaa4d23148f42ad58cccb5/src/main/scala/ai/eto/rikai/RikaiRelation.scala#L104
Saving to the path will go through this method, but saving as a table will not. I will take further investigating.

Renkai:
[Here](https://github.com/apache/spark/blob/v3.1.2/sql/core/src/main/scala/org/apache/spark/sql/execution/command/createDataSourceTables.scala#L216) should carry `__partition_columns` in  `table.storage.properties`, but it's not, will find out why.

```
ai.eto.rikai.RikaiOptions.<init>(RikaiOptions.scala:20)
ai.eto.rikai.RikaiRelationProvider.createRelation(RikaiRelationProvider.scala:68)
org.apache.spark.sql.execution.datasources.DataSource.writeAndRead(DataSource.scala:532)
org.apache.spark.sql.execution.command.CreateDataSourceTableAsSelectCommand.saveDataIntoTable(createDataSourceTables.scala:220)
[...]

PR #662, opened 2022-05-22
fix issue 640

Renkai:
This patch closes https://github.com/eto-ai/rikai/issues/640

Implemented `FileDataSourceV2` instead of `RelationProvider` to register a customized file format.

3 major diffs you may care:
* `metadata.json` is now saved before we save the dataframe, because I don't see a way to save it after we save the dataframe under the new API, and it seems fine to change the order
*  `RikaiOptions` has changed for being serializable, the new API need it serializable.
*  The default save old is not overwrite under new API, so we need to set save mode manually to keep safe.


python/rikai/spark/utils.py
```diff
@@ -22,6 +22,7 @@
 def df_to_rikai(df: "pyspark.sql.DataFrame", uri: str):
     (
         df.write.format("rikai")
+        .mode("overwrite")
```
eddyxu, line 25: why do we add `overwrite` everywhere?

Renkai: I think `overwrite` is the default behavior of the old save API. When we use the new API, the default behavior becomes `Error if exists`. And our test cases will reuse the same directory, so we need to declare it as `overwrite` explicitly to make the test case workable.

https://spark.apache.org/docs/latest/sql-data-sources-load-save-functions.html#save-modes

src/test/scala/ai/eto/rikai/RikaiRelationTest.scala
```diff
@@ -59,16 +63,39 @@ class RikaiRelationTest extends AnyFunSuite with SparkTestSession {
...
       Set(testDir.list().toSeq.filter(_.startsWith("label=")): _*)
 
     assert(partitions == Set("label=car", "label=people", "label=tree"))
   }
 
-  test("test default block size") {
-    val options = new RikaiOptions(Map.empty)
-    assert(options.blockSize == RikaiOptions.defaultBlockSize)
+  test("Use partitions in saveAsTable") {
+    import scala.reflect.io.Directory
```
darcy-shen, line 78: nit: `common-io` might be a better choice

da-tubi (approved):
LGTM

darcy-shen (approved):
LGTM

python/rikai/spark/utils.py
eddyxu: Isnt the unit test create a new temporary directory each time? It supposes to be different / empty directory for each individual test.

Also this `df_to_rikai` is not in test package, so it is dangerous to give the default value to `overwrite`?

Renkai: I will remove the non-test overwrite code and to see what will happen

Renkai: In `RikaiRelationTest` and several other tests, data would be saved to the same directory many times.
https://github.com/eto-ai/rikai/blob/94cd2d16c43f55266c0c548cad4f5839bdf20e9b/src/test/scala/ai/eto/rikai/RikaiRelationTest.scala#L26

Renkai: I think I understand what happened, in the old writer, it will check if the directory is empty, if it's empty then it's safe to write. The new writer will check if it exists, if it exists, the write will fail no matter the path is empty or not.

contrib/ros/tests/test_convert.py
```diff
@@ -111,7 +111,8 @@ def test_std_msgs_header(spark: SparkSession, tmp_path):
     )
     df.show()
     df.printSchema()
-    df.write.format("rikai").save(str(tmp_path))
+    df_path = str(tmp_path) + "df"
```
eddyxu, line 114: you can just use `df.write.format("rikai").save(str(tmp_path / "df"))`

python/tests/pytorch/test_torchvision.py
```diff
@@ -30,7 +30,9 @@ def test_vision_dataset(spark: SparkSession, tmp_path: Path):
     df = _create_dataframe(tmp_path, spark)
     df.show()
     dataset_dir = str(tmp_path / "data")
-    df.write.mode("overwrite").format("rikai").save(dataset_dir)
+    df.write.mode("overwrite").format("rikai").mode("overwrite").save(
```
eddyxu, line 33: Why does this have even more `overrite` ;p

Renkai: Fixed, the vanilla regex doesn't support semantic analysis, and I rely on it too much 😉

src/main/scala/ai/eto/rikai/RikaiRelationProvider.scala
```diff
@@ -35,7 +35,7 @@ class RikaiRelationProvider
     * spark.read.format("rikai").load("s3://path/to/featurestore")
     * }}}
     */
-  override def shortName(): String = "rikai"
+  override def shortName(): String = "rikaiold"
```
eddyxu, line 38: we should just delete this one if you feel confidence about the new format.

src/main/scala/ai/eto/rikai/RikaiFileFormat.scala
```diff
@@ -0,0 +1,60 @@
+package ai.eto.rikai
```
eddyxu, line 1: Apache license.

src/main/scala/ai/eto/rikai/RikaiFileFormat.scala
```diff
@@ -0,0 +1,60 @@
...
+    setSparkOptions(conf, rikaiOptions)
+
+    /** Rikai metadata directory. */
+    val rikaiDir = new Path(rikaiOptions.path, "_rikai")
+
+    /** Metadata file name */
+    val metadataFile = new Path(rikaiDir, "metadata.json")
+    writeMetadataFile(metadataFile, sparkSession, rikaiOptions)
+
+    super.prepareWrite(sparkSession, job, options, dataSchema)
```
eddyxu, line 58: can you confirm that `overwrite` will not overwrites the metadata files?

It seems the parquet fields were written AFTER metadata file

src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
```diff
@@ -1 +1,2 @@
-ai.eto.rikai.RikaiRelationProvider
\ No newline at end of file
+ai.eto.rikai.RikaiRelationProvider
```
eddyxu, line 1: We should just keep v2.

src/main/scala/ai/eto/rikai/RikaiDataSource.scala
```diff
@@ -0,0 +1,12 @@
+package ai.eto.rikai
```
eddyxu, line 1: Apache license.

python/tests/types/test_numpy.py
```diff
@@ -70,4 +70,4 @@ def resize_mask(arr: np.ndarray) -> np.ndarray:
     df = df.withColumn("resized", resize_mask("data"))
     df.show()
 
-    df.write.format("rikai").save(str(tmp_path))
+    df.write.format("rikai").mode("overwrite").save(str(tmp_path))
```
eddyxu, line 73: Can we use `df.write.format("rikai").save(str(tmp_path / "data"))` instead of specify overwrite manually in all files.

python/tests/spark/test_utils.py
```diff
@@ -25,6 +25,7 @@ def test_df_to_rikai(spark: SparkSession, tmp_path: Path):
     df = spark.createDataFrame(
         [Row(Box2d(1, 2, 3, 4)), Row(Box2d(23, 33, 44, 88))], ["bbox"]
     )
-    df_to_rikai(df, str(tmp_path))
-    actual_df = spark.read.format("rikai").load(str(tmp_path))
+    write_path = str(tmp_path) + "df"
```
eddyxu, line 28: `tmp_path / "df"`

src/main/scala/ai/eto/rikai/RikaiFileFormat.scala
Renkai: Do you mean if the metadata file would be deleted after the parquet files were written?

eddyxu (approved):

PR #667, opened 2022-06-03
Serialize width/height with polygon

eddyxu:


changhiskhan (approved):
Why is the width / height required for polygon? Is it just so we don't have to compute the bounding box on the fly?

PR #668, opened 2022-06-03
Fix FileRegistry missing base_dir, if the model URI is local path

eddyxu:


eddyxu:
The test failure is the same as the one being fixed in https://github.com/eto-ai/rikai/pull/667.

Due to a minor python version bump (3.9.12 -> 3.9.13) making CI not able to find cached packages. 

Will rebase after #667 is merged.

changhiskhan (approved):

PR #674, opened 2022-06-08
Fix bug converting array type

changhiskhan:


python/rikai/parquet/dataset.py
```diff
@@ -289,7 +289,9 @@ def _convert_col(self, col: pd.Series, field_type) -> pd.Series:
             return col.apply(partial(_convert_udt_value, udt=udt))
         elif field_type["type"] == "struct":
             return col.apply(lambda d: self._convert(d, field_type))
-        elif field_type["type"] == "array":
+        elif field_type["type"] == "array" and isinstance(
```
eddyxu, line 292: is there possible other case like `type == array and not isinstance(elementType, dict)`?

eddyxu, line 292: is there possible other case like `type == array and not isinstance(elementType, dict)`?

changhiskhan: if the elementType is not a dict then it's a simple native type which requires no conversion

eddyxu (approved):
I see. thanks for explanation

PR #687, opened 2022-06-22
Fix pandas writer issue

changhiskhan:
The pandas writer uses ToDict but Image.to_dict converted the bytes to string. When this string is then read back, the Image constructor thinks that it is actually the uri.

It looks like I originally made it convert to string, but I couldn't remember why. i searched through usages of `to_dict` and I didn't find any instance that really required the result to be non-binary.


eddyxu (approved):
LGTM

PR #693, opened 2022-07-05
Lazy torchvision model types

### LINKED ISSUE #692: Loading torchvision model types lazily
With #691, we will solve the problem that torchvision version is fixed.

However, for now, the torchvision dependencies are required even when we are using Rikai with Tensorflow.

https://github.com/eto-ai/rikai/blob/28726a0f8ffb1c9f092bbb7bb4987908b3472112/python/rikai/pytorch/models/__init__.py#L17-L26

I suggest that we should not load the above model types when torchvision is not installed.

da-tubi:
closes #692

### DIFF
--- a/python/rikai/pytorch/models/__init__.py
+++ b/python/rikai/pytorch/models/__init__.py
@@ -14,16 +14,21 @@
 
 """Rikai-implemented PyTorch models and executors."""
 
-import rikai.pytorch.models.convnext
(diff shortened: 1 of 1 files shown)


Renkai (approved):

darcy-shen:
The first CI run failed only for lint, the second CI run passed the lint and failed for other reasons.

PR #694, opened 2022-07-05
DSL to scale the Image

da-tubi:


python/rikai/types/vision.py
```diff
@@ -185,6 +185,20 @@ def display(self, **kwargs):
     def draw(self, drawable: Union[Drawable, list[Drawable], Draw]) -> Draw:
         return ImageDraw(self).draw(drawable)
 
+    def scale(self, factor: Union[int, float, Tuple]) -> Image:
+        if isinstance(factor, Tuple):
```
changhiskhan, line 189: is there any harm in accepting anything that's "list-like"? https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_list_like.html#pandas-api-types-is-list-like

changhiskhan (approved):
just the one nit

python/rikai/types/vision.py
darcy-shen: well, it is designed to use Tuple, and Tuple is immutable.

PR #697, opened 2022-07-08
pyspark use local[*] as master

da-tubi:


da-tubi:
There are 16 cores on my dev machine.

Renkai (approved):

da-tubi:
Using `local[*]` without increase the driver memory will result in a OOM exception.

PR #705, opened 2022-07-13
Update default Spark version to 3.2

changhiskhan:
Support matrix after PR:

Scala: 2.12, 2.13
Spark: 3.2
Python: 3.8, 3.9

Notes:
- Can't default to scala 2.13 because pyspark
- Python 3.7 is broken

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        scala-version: [ 2.12.10, 2.13.7 ]
-        java-version: ['8', '11']
+        scala-version: [ 2.13.8 ]
+        java-version: ['11']
     timeout-minutes: 15
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        scala-version: [ 2.13.8 ]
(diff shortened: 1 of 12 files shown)


build.sbt
```diff
@@ -2,8 +2,8 @@ import java.io.File
 import scala.reflect.io.Directory
 
 name := "rikai"
-crossScalaVersions := List("2.12.10", "2.13.7")
-scalaVersion := "2.12.10"
```
darcy-shen, line 6: the default scalaVersion should be 2.12.x

Because pyspark is default to using scala 2.12.x

python/rikai/spark/utils.py
```diff
@@ -72,7 +72,7 @@ def init_spark_session(
             "spark.jars.packages",
             ",".join(
                 [
-                    "ai.eto:rikai_2.12:{}".format(rikai_version),
+                    "ai.eto:rikai_2.1:{}".format(rikai_version),
```
darcy-shen, line 75: it should be 2.12, we do not need to change it

.github/workflows/scala.yml
```diff
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        scala-version: [ 2.12.10, 2.13.7 ]
-        java-version: ['8', '11']
+        scala-version: [ 2.12.16, 2.13.8 ]
+        java-version: ['11']
```
darcy-shen, line 15: we need java 8 because of https://docs.databricks.com/release-notes/runtime/10.4.html

darcy-shen (approved):
LGTM now

darcy-shen:
Please change the pull request title!

PR #707, opened 2022-10-18
Relax antlr python dependency

### LINKED ISSUE #706: Can we bump the version of `antlr4-python3-runtime`?
Hello!

Just asking: is there a reason that the version of `antlr4-python3-runtime` is pinned to `4.8` (https://github.com/eto-ai/rikai/blob/main/python/setup.py#L69)? I have seen multiple times that this conflicts with other libraries. Current stable version of that library is `4.11.1`.  I believe upgrading to at least `4.9` would help a lot.

eddyxu:
Closes #706


changhiskhan (approved):
LGTM

